PR #29, opened 2020-08-20
Consistent arg-name

tashrifbillah:
Testing and validation are different tasks for a neural network. Since the latter is being done during training, arguments that interact with the user should be appropriately named.


tashrifbillah:
By the way, just realized that you keep the validation data entirely separate, which emulates the definition of testing as well:
https://github.com/Deep-MI/FastSurfer/blob/c5e96778a237c7e21b1c525b8cf8cf1fc95cfdf2/FastSurferCNN/models/solver.py#L360

I was under the impression that validation data are also used in training.

Nevertheless, the proposed change holds for clarity and consistency.

LeHenschel:
I agree, consistent naming is important. I already changed this a while back so I am going to close the pull request. Thanks!

PR #58, opened 2021-08-03
conform does not handle 4-dimensional data frames correctly, if the fourth dim is size 1

dkuegler:
Fix an error that occurs when the file passed to conform that has frames (conform does not handle 4-dimensional data frames correctly, if the fourth dim is size 1)


dkuegler:
I updated to more conservative approach

dkuegler:
I also double checked this implementation with mri_convert --conform results and results are sound.

PR #72, opened 2021-10-29
changed np.float to float

VellaRo:
changed np.float to float in all files


LeHenschel:
Please pull the new version and then push the float changes.

PR #75, opened 2021-10-31
changed example with --data_dir to subject1 till subject19

VellaRo:
Changed the example from subject1 to subject1-19 so it is clearer that you can adjust the selection of the subjects.


FastSurferCNN/README.md
```diff
@@ -131,6 +131,21 @@ python3 generate_hdf5.py \
 
 ```
 
+#### Example Command Sagittal using --data_dir instead of --csv_file
+--data_dir specifies the path in which the data is located, with --pattern we can select subjects from the specified path. By default the pattern is "*" meaning all subjects will be selected.
+In the following example we selecting subject1 till subject19.
```
m-reuter, line 136: ... we select subject1 till subject9.

FastSurferCNN/README.md
```diff
@@ -131,6 +131,21 @@ python3 generate_hdf5.py \
...
+```
+python3 generate_hdf5.py \
+--hdf5_name ../data/training_set_cispa_axial.hdf5 \
+--data_dir ../data \
+--pattern "subject1[0-9]" \
```
m-reuter, line 141: this looks like you are selecting subject10 to subject19 ? It should match the description above.

FastSurferCNN/README.md
LeHenschel: Change this line to subject10 to subject19 (which is what we select with --pattern "subject1[0-9]".

LeHenschel (changes requested):
I think it would be better to make this a bit more descriptive. After the sentence "By default...", we could start with an example: 

As an example, imagine you have 19 FreeSurfer processed subjects labeled subject1 to subject19 in the ../data directory:
```bash
/home/user/FastSurfer/data
├── subject1
├── subject2
├── subject3
…
│
├── subject19
    ├── mri
    │   ├── aparc.DKTatlas+aseg.mgz
    │   ├── aseg.auto_noCCseg.mgz
    │   ├── orig.mgz
    │   ├── …
    │   …
    ├── scripts
    ├── stats
    ├── surf
    ├── tmp
    ├── touch
    └── trash
```  
, setting --pattern "*" will select all 19 subjects (subject1, ..., subject19). Now, if only a subset should be used for the hdf5-file (e.g. subject 10 till subject19), this can be done by changing the --pattern flag to "subject1[0-9]": example command...

This way it will be clear what "all" means. It also highlights the subject directory structure.

PR #81, opened 2021-11-09
Add requirements.in, and pip-compiled requirements.txt

af-a:
## Description

This PR swaps the manually-written `requirements.txt` file for the version generated using the `pip-compile` tool.

A `requirements.in` file containing only the direct dependencies used in the package is added. The resultant `requirements.txt` file lists all sub-dependencies of each package.


LeHenschel (commented):
The versions of the secondary dependencies are now different than before. I think it would be good to have equivalent version across Docker, yaml and requirements.txt

requirements.txt
```diff
@@ -1,34 +1,68 @@
-cycler==0.10.0
+#
+# This file is autogenerated by pip-compile with python 3.6
+# To update, run:
+#
+#    pip-compile --output-file=-
+#
+--find-links https://download.pytorch.org/whl/torch_stable.html
+
+cycler==0.11.0
```
LeHenschel, line 9: cycler 0.10.0 changed to 0.11.0

AhmedFaisal95: Strangely, `pip-compile` **sometimes** introduced the different package versions when run multiple times in the same environment on the same `requirements.in` file.

I've pushed a correctly-generated `requirements.txt` file in https://github.com/Deep-MI/FastSurfer/pull/81/commits/d6298868ca2e03b16d1ba68a2f00b8f538cc169b

LeHenschel (approved):

LeHenschel:
For future reference, pip-compile is part of pip-tools and hence must be installed via pip:

`python3 -m pip install --user pip-tools`

Usage description from [pypi-piptools](https://pypi.org/project/pip-tools/)

> Example usage for pip-compile
> The pip-compile command lets you compile a requirements.txt file from your dependencies, specified in either setup.py or requirements.in.
> 
> Run it with pip-compile or python -m piptools compile. If you use multiple Python versions, you can also run py -X.Y -m piptools compile on Windows and pythonX.Y -m piptools compile on other systems.
> 
> pip-compile should be run from the same virtual environment as your project so conditional dependencies that require a specific Python version, or other environment markers, resolve relative to your project’s environment.
> 
> Note: If pip-compile finds an existing requirements.txt file that fulfils the dependencies then no changes will be made, even if updates are available. To compile from scratch, first delete the existing requirements.txt file, or see [Updating requirements](https://pypi.org/project/pip-tools/#updating-requirements) for alternative approaches.

PR #86, opened 2021-11-30
Saving FastSurferCNN output in mgz, nii, or nii.gz

AhmedFaisal95:
## Description

This PR incorporates saving of FastSurferCNN output in mgz, nii, or nii.gz formats, through the `save_image` function.

The desired output format is inferred from the output file name, and the function handles saving the output in that format.
Note that an `AssertionError` is raised for an invalid or unsupported file format.

Note: the default segmentation output format has been set to nii.gz (the documentation has also been updated to reflect this).

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -16,7 +16,7 @@ In order to build the docker image for FastSurfer (FastSurferCNN + recon-surf; o
 ./docker_build.sh
 ```
 
-This script builds a docker image with the name fastsurfer:gpu. With it you basically execute the script __run_fastsurfer.sh__ from the parent directory. It takes as input a single T1-weighted MRI brain scan (from the /data directory) and first produces the aparc.DKTatlas+aseg.mgz segmentation followed by the surface construction (output stored in /output directory).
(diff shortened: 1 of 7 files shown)


LeHenschel:
I think it is not a good idea to just change the default to .nii.gz. We should at least add a check for .mgz if the default .nii.gz is not found in run_fastsurfer.sh. We have been using .mgz for quite some time and this might mess with processing steps other users have already set up. I know that the switch is simple (the user can just provide the additional flag to use .mgz), but I would either keep the default output to .mgz or have an automatic check if the .nii.gz is not found. @m-reuter, what do you think?

LeHenschel:
After discussing the issue with Martin, we decided to keep the default format at .mgz for the time being. Instead, a separate flag should be issued if the user wants nifti-output (or automatically detect ending in output segmentation file name).

AhmedFaisal95:
I have reverted the relevant changes; the default format is mgz as before.
The desired output format is still inferred from the output filename extension.

I have made small additions to the READMEs to mention the supported segmentation output file formats. Let me know if you prefer this be documented differently.

LeHenschel (changes requested):
For the save_image functionality, I think it would be better to put this in the data_loader/load_neuroimaging_data.py as it might be useful in other settings as well (e.g. for changing the output file format in recon-surf with conform.py).

PR #87, opened 2021-11-30
Singularity readme

VellaRo:
Readme equivalent to Docker/README.md: (First Version)

*Singularity setup

* the whole FastSurfer pipeline (FastSurferCNN + recon-surf, Example 1 (GPU) Example 2 (CPU)) (Tested)
* only the segmentation network (FastSurferCNN, Example 3 (GPU) Example 4 (CPU)) (Tested)
* only the surface pipeline (recon-surf, Example 5 (CPU)) (Tested)

*Frequent Problems List


LeHenschel (changes requested):
Couple of points we should address:

- [ ] The description is very detailed (token generation etc.). I think it is enough to refer to the relevant pages where this is described. Here, we should focus on how to run FastSurfer singularity, not singularity in general. You can add this item to the FAQ at the end
- [ ] It is not clear to me what the tmp_data is for. Is this the place where my files will be stored or the input directory? Can I bind a directory only for data input or also data output? Would be good to give an example where what is stored/taken from
- [ ] Multiple bash commands following each other can be grouped into one
- [ ] I would first give the run command for the singularity container and afterwards explain the flags
- [ ] Some sections are very repetitive (export license...). Maybe this can be grouped at the beginning (like initial set up)
- [ ] I like the idea of the FAQ. I think it would be more helpful to change the wording so the user can find the solution faster. Normally they will just see an error and search for it (e.g. "I see error message XYZ" Answer: Export the license...; "It does not do XYZ" Answer: check ABC...)
- [ ] There are some spelling mistakes (u instead of you, he instead of the, freesurfer instead of FreeSurfer)

m-reuter:
Currently we are updating the docker images with the goal to make them weigh less. 
Furthermore, we currently recommend people to build docker images and then convert them into singularity. 
Therefore this will be closed and we will re-visit singularity later.

PR #88, opened 2021-11-30
Singularity

VellaRo:
Singularity defenition files for:
* the whole FastSurfer pipeline (FastSurferCNN + recon-surf,  (GPU)  (CPU))
* only the segmentation network (FastSurferCNN, (GPU)  (CPU))
* only the surface pipeline (recon-surf,  (CPU))

All tested with --remote build option.


LeHenschel (commented):
Looks good to me. Can you specify which versions (flag combinations) you tested and on how many subjects?

VellaRo:
I tested it on only one subjects at a time with the same flags as in the Examples 1 - 5. I will run some more tests with multiple subjects and different flags combination.

m-reuter:
Currently we are updating the docker images with the goal to make them weigh less. 
Furthermore, we currently recommend people to build docker images and then convert them into singularity. 
Therefore this will be closed and we will re-visit singularity later.

Issue #45, opened 2021-03-19
Crash in nu_correct if FOV > 256


davecash75:
Hi there-
I'm currently using the CPU Docker and I am running the standard docker command for the full FastSurfer as you have in the README, with the exception that I pass in a Nifti where the FOV is greater than 256 mm. WHen I run, the deep_seg runs fine but during the recon, I get a crash with the following key messages: 
WARNING ==================++++++++++++++++++++++++=======================================
The physical sizes are (228.80 mm, 282.00 mm, 282.00 mm), which cannot fit in 256^3 mm^3 volume.
The resulting volume will have 282 slices.
If you find problems, please let us know (freesurfer@nmr.mgh.harvard.edu).
==================================================++++++++++++++++++++++++===============
Mask volume and input volume must be the same size.
nu_evaluate: crashed while running evaluate_field (termination status=512)
nu_correct: crashed while running nu_evaluate (termination status=512)
ERROR: nu_correct
I run mri_convert with the --cw256 command outside of FastSurfer, and it no longer has this error. So I guess the question is would it be possible to have an option that conforms (and crops) images with FOV > 256 mm just like the original FreeSurfer allows.
Many thanks,
David Cash

m-reuter:
Hi David, yes such a flag would make sense. Currently, with large field of view, you need to convert images yourself before you pass them into FastSurfer. If you end up modifying our scripts to take such a flag, please generate a pull request. Thanks

m-reuter:
I have looked into this a little. conform.py seems to be able to automatically conform these to 255 cube images. So not sure why nu_correct fails later. It could be that the stream does not use this image, but instead runs a second conform (potentially from freeSurfer) without the -cw256 flag at a later stage again. We need to check the pipeline steps.

m-reuter:
should be fixed by #92

PR #92, opened 2021-12-21
Saving conformed input in FastSurferCNN to use in recon-surf

AhmedFaisal95:
## Description

This PR adds saving of the conformed image used as input to the FastSurferCNN network, so that it can later be used in `recon-surf.sh`. 

Previously, `mri_convert  -c` was used in `recon-surf.sh` to create `orig.mgz` from the input image. If the input image was not conformed to 256 beforehand, this caused a mismatch between the image and output segmentation (leading to issues as in  https://github.com/Deep-MI/FastSurfer/issues/45).

The modification ensures that the mismatch does not occur by always passing in the saved conformed input image to `recon-surf.sh` (through `run_fastsurfer.sh` by default).
The `conform.py` script is used to check whether both input image and segmentation are conformed in `recon_surf.sh` (for the latter, the UCHAR criterion is disregarded). If either is not, `recon_surf.sh` will fail. This check is necessary to validate inputs that are not produced by FastSurferCNN, for example.
The script has been slightly modified to enable this behaviour.

### Changes

Here's a brief summary of the changes:
 - `is_conform` also checks whether `dtype` of input is UCHAR (`uint8`)
 - `is_conform` can display specifically which conformance criteria were violated (controlled by verbosity flag: off by default in the script, but on by default in `recon_surf.sh`); see example output below
 - `is_conform` has a `check_dtype` flag, which is off when the input image is a segmentation
 - `conform.py` has `check_only` (only check if the input image is conformed) and `seg_input` (consider the input image a segmentation, and set `check_dtype` to `True`) flags, and returns appropriate error codes
 - `eval.py` writes out the conformed input image
 - `run_fastsurfer.sh` specifies `conformed_input`, the path of the to-be conformed input image (default: `/mri/conformed_orig.mgz`)
 - `run_fastsurfer.sh` passes in `conformed_input` to the `t1` argument of `recon-surf.sh` by default
 - `recon_surf.sh` uses `conform.py` to check for conformance of `t1` and `seg`

### `is_conform` verbose output example

An example of a non-conformed input image being used as input to `conform.py`, with flags `check_only` and `verbose`:

```
Reading input: ./001.mgz ...
The input image is not conformed.
A conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        False
 - Voxel size 1x1x1:              False
 - Orientation LIA:               False
 - Dtype uint8:                   False
check_only flag provided. Exiting without conforming input image.
```

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -55,7 +55,8 @@ def options_parse():
     parser = optparse.OptionParser(version='$Id: conform.py,v 1.0 2019/07/19 10:52:08 mreuter Exp $',
                                    usage=HELPTEXT)
     parser.add_option('--input', '-i', dest='input', help=h_input)
-    parser.add_option('--output', '-o', dest='output', help=h_output)
+    parser.add_option('--output', '-o', dest='output', help=h_output,
+                      default='./conformed_img.mgz')
     parser.add_option('--order', dest='order', help=h_order, type="int", default=1)
     parser.add_option('--check_only', dest='check_only', default=False, action='store_true',
                       help='If True, only checks if the input image is conformed')
@@ -63,8 +64,8 @@ def options_parse():
                       help='Specifies whether the input is a seg image. If true, '
                            'the check for conformance disregards the uint8 dtype criteria')
     (fin_options, args) = parser.parse_args()
-    if fin_options.input is None or fin_options.output is None:
-        sys.exit('ERROR: Please specify input and output images')
+    if fin_options.input is None:
+        sys.exit('ERROR: Please specify input image')
     return fin_options
 
 
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -57,6 +57,11 @@ def options_parse():
     parser.add_option('--input', '-i', dest='input', help=h_input)
     parser.add_option('--output', '-o', dest='output', help=h_output)
     parser.add_option('--order', dest='order', help=h_order, type="int", default=1)
+    parser.add_option('--check_only', dest='check_only', default=False, action='store_true',
+                      help='If True, only checks if the input image is conformed')
+    parser.add_option('--seg_input', dest='seg_input', default=False, action='store_true',
+                      help='Specifies whether the input is a seg image. If true, '
+                           'the check for conformance disregards the uint8 dtype criteria')
     (fin_options, args) = parser.parse_args()
     if fin_options.input is None or fin_options.output is None:
         sys.exit('ERROR: Please specify input and output images')
@@ -362,8 +367,19 @@ if __name__ == "__main__":
     if len(image.shape) > 3 and image.shape[3] != 1:
         sys.exit('ERROR: Multiple input frames (' + format(image.shape[3]) + ') not supported!')
 
-    if is_conform(image):
-        sys.exit("Input " + format(options.input) + " is already conform! No output created.\n")
+    if not options.seg_input:
+        image_is_conformed = is_conform(image, check_dtype=True)
+    else:
+        image_is_conformed = is_conform(image, check_dtype=False)
+
+    if image_is_conformed:
+        print("Input " + format(options.input) + " is already conformed! Exiting.\n")
+        sys.exit(0)
+    else:
+        # Note: if check_only, a non-conforming image leads to an error code, this result is needed in recon_surf.sh
+        if options.check_only:
+            print("check_only flag provided. Exiting without conforming input image.\n")
+            sys.exit(1)
 
     # If image is nifti image
     if options.input[-7:] == ".nii.gz" or options.input[-4:] == ".nii":
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -63,6 +63,8 @@ def options_parse():
     parser.add_option('--seg_input', dest='seg_input', default=False, action='store_true',
                       help='Specifies whether the input is a seg image. If true, '
                            'the check for conformance disregards the uint8 dtype criteria')
+    parser.add_option('--verbose', dest='verbose', default=False, action='store_true',
+                      help='If verbose, more specific messages are printed')
     (fin_options, args) = parser.parse_args()
     if fin_options.input is None:
         sys.exit('ERROR: Please specify input image')
@@ -256,7 +258,7 @@ def conform(img, order=1):
     return new_img
 
 
-def is_conform(img, eps=1e-06, check_dtype=True):
+def is_conform(img, eps=1e-06, check_dtype=True, verbose=True):
     """
     Function to check if an image is already conformed or not (Dimensions: 256x256x256, Voxel size: 1x1x1, and
     LIA orientation.
@@ -299,9 +301,11 @@ def is_conform(img, eps=1e-06, check_dtype=True):
     if all(criteria.values()):
         return True
     else:
-        print('The input image is not conformed. A conformed image must satisfy the following criteria:')
-        for condition, value in criteria.items():
-            print(' - {:<30} {}'.format(condition+':', value))
+        print('The input image is not conformed.')
+        if verbose:
+            print('A conformed image must satisfy the following criteria:')
+            for condition, value in criteria.items():
+                print(' - {:<30} {}'.format(condition+':', value))
         return False
 
 
@@ -369,9 +373,9 @@ if __name__ == "__main__":
         sys.exit('ERROR: Multiple input frames (' + format(image.shape[3]) + ') not supported!')
 
     if not options.seg_input:
-        image_is_conformed = is_conform(image, check_dtype=True)
+        image_is_conformed = is_conform(image, check_dtype=True, verbose=options.verbose)
     else:
-        image_is_conformed = is_conform(image, check_dtype=False)
+        image_is_conformed = is_conform(image, check_dtype=False, verbose=options.verbose)
 
     if image_is_conformed:
         print("Input " + format(options.input) + " is already conformed! Exiting.\n")
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -250,7 +250,7 @@ def conform(img, order=1):
     return new_img
 
 
-def is_conform(img, eps=1e-06):
+def is_conform(img, eps=1e-06, check_dtype=True):
     """
     Function to check if an image is already conformed or not (Dimensions: 256x256x256, Voxel size: 1x1x1, and
     LIA orientation.
@@ -279,6 +279,10 @@ def is_conform(img, eps=1e-06):
     iaffine = img.affine[0:3, 0:3] + [[1, 0, 0], [0, 0, -1], [0, 1, 0]]
 
     if np.max(np.abs(iaffine)) > 0.0 + eps:
+        return False
+
+    # check dtype uchar
+    if check_dtype and img.get_data_dtype() != 'uint8':
         return False
 
     return True
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -261,6 +261,9 @@ def is_conform(img, eps=1e-06, check_dtype=True):
                       thus sometimes not correctly recognized. The epsilon accounts for these small shifts.
     :return: True if image is already conformed, False otherwise
     """
+
+    criteria = {'Dimensions 256x256x256': True, 'Voxel size 1x1x1': True, 'Orientation LIA': True}
+
     ishape = img.shape
 
     if len(ishape) > 3 and ishape[3] != 1:
@@ -268,24 +271,32 @@ def is_conform(img, eps=1e-06, check_dtype=True):
 
     # check dimensions
     if ishape[0] != 256 or ishape[1] != 256 or ishape[2] != 256:
-        return False
+        criteria['Dimensions 256x256x256'] = False
 
     # check voxel size
     izoom = img.header.get_zooms()
     if izoom[0] != 1.0 or izoom[1] != 1.0 or izoom[2] != 1.0:
-        return False
+        criteria['Voxel size 1x1x1'] = False
 
     # check orientation LIA
     iaffine = img.affine[0:3, 0:3] + [[1, 0, 0], [0, 0, -1], [0, 1, 0]]
 
     if np.max(np.abs(iaffine)) > 0.0 + eps:
-        return False
+        criteria['Orientation LIA'] = False
 
     # check dtype uchar
-    if check_dtype and img.get_data_dtype() != 'uint8':
-        return False
+    if check_dtype:
+        criteria['Dtype uint8'] = True
+        if img.get_data_dtype() != 'uint8':
+            criteria['Dtype uint8'] = False
 
-    return True
+    if all(criteria.values()):
+        return True
+    else:
+        print('The input image is not conformed. A conformed image must satisfy the following criteria:')
+        for condition, value in criteria.items():
+            print(' - {:<30} {}'.format(condition+':', value))
+        return False
 
 
 def check_affine_in_nifti(img, logger=None):
--- a/FastSurferCNN/eval.py
+++ b/FastSurferCNN/eval.py
@@ -88,6 +88,8 @@ def options_parse():
                         help='name under which segmentation will be saved. Default: aparc.DKTatlas+aseg.deep.mgz. '
                              'If a separate subfolder is desired (e.g. FS conform, add it to the name: '
                              'mri/aparc.DKTatlas+aseg.deep.mgz)')
+    parser.add_argument('--conformed_input', '--conformed_input', dest='conformed_input', default='conformed_input.mgz',
+                        help='Name under which the conformed input image will be saved. Default: conformed_input.mgz.')
     parser.add_argument('--order', dest='order', type=int, default=1,
                         help="order of interpolation (0=nearest,1=linear(default),2=quadratic,3=cubic)")
 
@@ -265,6 +267,11 @@ def fastsurfercnn(img_filename, save_as, use_cuda, gpu_small, logger, args):
     logger.info("Reading volume {}".format(img_filename))
 
     header_info, affine_info, orig_data = load_and_conform_image(img_filename, interpol=1, logger=logger)
+
+    # Save conformed input image to use in recon-surf later
+    conformed_orig_img = nib.MGHImage(orig_data, affine_info, header_info)
+    conformed_orig_img.to_filename(args.conformed_input)
+    logger.info("Saving conformed original image to {}".format(args.conformed_input))
 
     # Set up model for axial and coronal networks
     params_network = {'num_channels': args.num_channels, 'num_filters': args.num_filters,
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -413,10 +413,10 @@ echo "================== Creating orig and rawavg from input ===================
 echo " " |& tee -a $LF
 
 # check for input conformance
-cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $t1 --check_only"
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $t1 --check_only --verbose"
 RunIt "$cmd" $LF
 
-cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $seg --check_only --seg_input"
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $seg --check_only --seg_input --verbose"
 RunIt "$cmd" $LF
 
 # create orig.mgz and aparc+aseg.orig.mgz (copy of segmentation)
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -412,8 +412,15 @@ echo " " |& tee -a $LF
 echo "================== Creating orig and rawavg from input =========================" |& tee -a $LF
 echo " " |& tee -a $LF
 
+# check for input conformance
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $t1 --check_only"
+RunIt "$cmd" $LF
+
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $seg --check_only --seg_input"
+RunIt "$cmd" $LF
+
 # create orig.mgz and aparc+aseg.orig.mgz (copy of segmentation)
-cmd="mri_convert -c $t1 $mdir/orig.mgz"
+cmd="mri_convert $t1 $mdir/orig.mgz"
 RunIt "$cmd" $LF
 
 cmd="mri_convert $seg $mdir/aparc+aseg.orig.mgz"
--- a/run_fastsurfer.sh
+++ b/run_fastsurfer.sh
@@ -273,6 +273,11 @@ if [ -z "$seg" ]
     seg="${sd}/${subject}/mri/aparc.DKTatlas+aseg.deep.mgz"
 fi
 
+if [ -z "$conformed_input" ]
+  then
+    conformed_input="${sd}/${subject}/mri/conformed_orig.mgz"
+fi
+
 if [ -z "$seg_log" ]
  then
     seg_log="${sd}/${subject}/scripts/deep-seg.log"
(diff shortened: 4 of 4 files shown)


LeHenschel (changes requested):
There are some issues which need to be addressed. Also please test all variations (e.g. what happens  if I do not define the flags (defaults are used), what happens if I run FastSurferCNN alone then again recon-surf alone with the run_fastsurfer.sh command and identical input arguments, input conformed/non-conformed image and different file formats ...):

**conform.py:**

- [ ] 1. Should not have a default. Flag should always be passed via calling scripts (e.g. eval.py) 
- [ ] 2. Also currently the default name here is **conformed_img.mgz** which differs from default names in the other scripts (eval and run_fastsurfer.sh). All should be changed to mri/orig.mgz

**eval.py:**
- [ ] 1. Flag should not be called conformed_input but conformed_output (the input is not conformed, the output will be)
- [ ] 2. The default here is **conformed_input.mgz** which differs from the one specified in conform.py. Default should be adapted to mri/orig.mgz (same location the segmentation is saved) based on the FreeSurfer convention)
- [ ] 4. Flag description should indicate that the input image will be conformed if it is not and that this version is saved under the given location x
- [ ] 5. In accordance with FreeSurfer, we should also save the original image by default in a subdirectory called “orig” of the specified output directory (i.e. if output segmentation will be saved in mri/aparc.DKTatlas+aseg.deep.mgz, the conformed image will be stored in mri/orig.mgz and the original one in mri/orig/001.mgz)
- [ ] 6. For convenience, we should make sure that the image is saved in the same format requested as the output segmentation (so .mgz, .nii or nii.gz)
- [ ] 7. Given, that a saving routine now exists, II think it makes sense to use it for storing the conformed and original output as well instead of the explicit nib.MGHImage … to_filename route

**recon-surf.sh**
- [ ]  1. The documentation should be updated that --t1 needs to be a conformed image now. Here, we might also point out how to conform the input image (conform.py)
- [ ]  2. Same for the segmentation (update documentation, mention that FastSurferCNN does this by default but if you use another one, need to conform it; mention downside of resampling segmentation)

**run_fastsurfer.sh**
- [ ] 1. The flag “—conformed_input” does not exist. There is no way for the user to overwrite it. Documentation is missing as well
- [ ] 2. The default is different from eval.py and conform.py (**mri/conformed_orig.mgz**). It should be changed to mri/orig.mgz (same note as for eval.py)


**READMEs**:
- [ ] The READMEs need to be updates to reflect the changes (describe new flags, input to recon-surf should be conformed...)

m-reuter:
about conform.py I do not think the script itself should have a default filename in it anywhere! It should always be passed as a parameter.

m-reuter:
recon_surf   check_only should not convert images (as it is a check routine and should not change things). We can keep using mri_convert for that for now. Or we can write our own convert.py to cover some of the mri_convert functionality, e.g. converting between image formats. For now it is sufficient to keep using mri_convert (without the -c flag of course, only converting nil to mgz for example).

m-reuter:
And finally run_fastsurfer : is this a wrapper around everything (including segmentation)? Then why does the input need to be a conformed images? It should conform it by default. Only recon_surf will require conform images together with conform segmentation to work. Most people will not call recon_surf directly. The supported use cases are: 
- call everything (including segmentation and surface processing)
- call only segmentation (here in the future we need to add a simple way of producing volume statistics similar to aseg.stats directly for users who do not care about surfaces, but would like to use the stats from the segmentation).

LeHenschel:
I updated the list. The --t1 and --seg conform requirement should of course be for recon-surf.sh, not run_fastsurfer.sh (added to the wrong subsection). Personally, I think it is more convenient to directly return the .mgz image in check_conform when a specific run_fastsurfer flag or sth is set. But we can also keep mri_convert. Potentially, we can also just use the save_img routine to do this (as it is just a matter of changing the file-type).

One note, with run_fastsurfer.sh you can also just run the surface pipeline. This is useful, if e.g. you have a GPU server and CPU server. So you first generate the segmentations only on the GPU and then run the recon-surf part on  a CPU server. We have to make sure that in this case, the same command will run through both times (automatically finding the conformed image).

AhmedFaisal95:
Thanks for the feedback @LeHenschel.
I'll update the PR accordingly. Regarding testing, I usually only push changes after having extensively tested on different case variations; I'll make sure to cover all instances I can think of.

To clarify a few points:
* I was going to include converting to mgz in `conform.py`, but I recalled @m-reuter mentioning at some point that he did not want `conform.py` to perform any conversions
* `conformed_input` (name subject to improvement) was not exposed as an argument in `run_fastsurfer.sh`, because I thought we would want the filename to be standardized/fixed (e.g. always mri/orig.mgz). But I’ll make that parameterizable too.

AhmedFaisal95:
## Update

The following changes have been made:
 * removing default output filename in `conform.py`
 * in `conform.py`, if `--check_only` and an output filename is given, an error is raised
 * renaming `conformed_input` to `conformed_name` (following discussions)
 * exposing `run_fastsurfer.sh` argument `conformed_name`
 * the default value of `conformed_name`: `${sd}/${subject}/mri/orig.mgz` 
 * saving of original image in `${sd}/${subject}/mri/orig/001.mgz` (in `eval.py`); the method used here will be changed once the PR containing the `save_image` functionality is merged (note that this is currently always saved in `mgz` format)
 * updating documentation: descriptions of arguments in respective scripts
 * updating READMEs with mentions of `conformed_name` argument, and requirements of conformed image and segmentation inputs to `recon_surf` (including note on possibility of utilizing `conform.py` for non-conformed images).

### Still Open
These should address the items in the list above, with the exception of the following:
* Items 6 and 7 under `eval.py` will be addressed once https://github.com/Deep-MI/FastSurfer/pull/86 is merged; `save_image` contains the functionality needed to save the image in one of the supported formats.
* In item 2 under `recon-surf.sh`, a mention of the "downside of resampling segmentation" is still missing. I am not clear on this yet.

### Testing
For more explicit details on testing, I'm also including the different test cases (combinations of inputs and command variations) that I used to validate the modifications to i) using the conformed image within the pipeline, ii) `conform.py` (with the new flags, etc.) below.

For `conform.py`, the terminal output is also included in each case.

<details>
<summary> Pipeline tests </summary>


These test cases include the following variations:
* Whether the input was conformed or not
* Which parts of the pipeline were run (full, seg+surf with main script, seg+surf with separate scripts)
* Whether the input was a mgz, nii, or nii.gz file (last two in cases 7 and 8, respectively)

#### 1. Running full pipeline with `run_fastsurfer.sh` on conformed input image
Command:
```
/run_fastsurfer.sh --t1 /data_dir/mri/T1.mgz --sd /output_dir/test_case_1 --sid 00001 --batch 8 --parallel --fs_license /fs60/.license
```
Result: Succeeds, as expected.

#### 2. Running segmentation and surface generation separately with `run_fastsurfer.sh` on conformed input image
Commands:
```
./run_fastsurfer.sh --t1 /data_dir/mri/T1.mgz --sd /output_dir/test_case_2 --sid 00001 --batch 8 --seg_only
```
```
./run_fastsurfer.sh --t1 /output_dir/test_case_2/00001/mri/orig.mgz --sd /output_dir/test_case_2 --sid 00001 --parallel --fs_license /fs60/.license --surf_only
```
Result: Succeeds, as expected.

#### 3. Running segmentation and surface generation separately with `eval.py` and `recon-surf.sh` on conformed input image
Commands:
```
python3.6 ./FastSurferCNN/eval.py --in_name /data_dir/mri/T1.mgz --out_name /output_dir/test_case_3/00001/mri/aparc.DKTatlas+aseg.deep.mgz --conformed_name /output_dir/test_case_3/00001/mri/orig.mgz --batch_size 8 --simple_run
```
```
./recon_surf/recon-surf.sh --t1 /output_dir/test_case_3/00001/mri/orig.mgz --sd /output_dir/test_case_3 --sid 00001 --parallel --fs_license /fs60/.license
```
Result: Succeeds, as expected.

#### 4. Running full pipeline with `run_fastsurfer.sh` on non-conformed input image
Command:
```
./run_fastsurfer.sh --t1 /data_dir/mri/orig/001.mgz --sd /output_dir/test_case_4 --sid 00001 --batch 8 --parallel --fs_license /fs60/.license
```
Result: Succeeds, as expected.


#### 5. Surface generation with `recon_surf.sh` on non-conformed input image, conformed input segmentation
(NOTE: `eval.py` is just used to generate the segmentation)
Commands:
```
python3.6 ./FastSurferCNN/eval.py --in_name /data_dir/mri/orig/001.mgz --out_name /output_dir/test_case_5/00001/mri/aparc.DKTatlas+aseg.deep.mgz --conformed_name /output_dir/test_case_5/00001/mri/orig.mgz --batch_size 8 --simple_run
```
```
./recon_surf/recon-surf.sh --t1 /data_dir/mri/orig/001.mgz --sd /output_dir/test_case_5 --sid 00001 --parallel --fs_license /fs60/.license
```
Result: `recon_surf.sh` fails due to the non-conformed image, as expected, with the following error:
```
python3.6 /fastsurfer/recon_surf/../FastSurferCNN/data_loader/conform.py -i /data_dir/mri/orig/001.mgz --check_only --verbose
Reading input: /data_dir/mri/orig/001.mgz ...
The input image is not conformed.
A conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        False
 - Voxel size 1x1x1:              False
 - Orientation LIA:               False
 - Dtype uint8:                   False
check_only flag provided. Exiting without conforming input image.

```

#### 6. Surface generation with `recon_surf.sh` on conformed input image, non-conformed input segmentation
(NOTE: `eval.py` is used to generate a "faulty" segmentation. This segmentation is manipulated to violate one of the conformance criteria by adding an extra dimension in the middle channel using `pred_prob = np.hstack((pred_prob, np.zeros((pred_prob.shape[0], 1, pred_prob.shape[2]))))`. The `seg` argument is then used to pass in this segmentation as input to `recon_surf.sh`)
Commands:
```
python3.6 ./FastSurferCNN/eval.py --in_name /data_dir/mri/T1.mgz --out_name /output_dir/test_case_6/00001/mri/aparc.DKTatlas+aseg.deep.faulty.mgz --conformed_name /output_dir/test_case_6/00001/mri/orig.mgz --batch_size 8 --simple_run
```
```
./recon_surf/recon-surf.sh --t1 /output_dir/test_case_6/00001/mri/orig.mgz --sd /output_dir/test_case_6 --sid 00001 –seg /output_dir/test_case_6/00001/mri/aparc.DKTatlas+aseg.deep.faulty.mgz --parallel --fs_license /fs60/.license
```
Result: `recon_surf.sh` fails due to the non-conformed segmentation, as expected, with the following error:
```
python3.6 /fastsurfer/recon_surf/../FastSurferCNN/data_loader/conform.py -i /output_dir/test_case_6/00001/mri/aparc.DKTatlas+aseg.deep.faulty.mgz --check_only --seg_input --verbose
Reading input: /output_dir/test_case_6/00001/mri/aparc.DKTatlas+aseg.deep.faulty.mgz ...
The input image is not conformed.
A conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        False
 - Voxel size 1x1x1:              False
 - Orientation LIA:               True
check_only flag provided. Exiting without conforming input image.

```

#### 7. Running full pipeline with `run_fastsurfer.sh` on non-conformed input image in `nii` format
(NOTE: the input was generated using `mri_convert 001.mgz 001.nii`)
Command:
```
./run_fastsurfer.sh --t1 /fastsurfer/001.nii --sd /output_dir/test_case_7 --sid 00001 --batch 8 --parallel --fs_license /fs60/.license
```
Result: Succeeds, as expected.

#### 8. Running full pipeline with `run_fastsurfer.sh` on non-conformed input image in `nii.gz` format
(NOTE: the input was generated using `mri_convert 001.mgz 001.nii.gz`)
Command:
```
./run_fastsurfer.sh --t1 /fastsurfer/001.nii.gz --sd /output_dir/test_case_8 --sid 00001 --batch 8 --parallel --fs_license /fs60/.license
```
Result: Succeeds, as expected.
</details>

<details>
<summary> conform.py tests </summary>


These test cases include the following variations:
* Whether the input was conformed or not
* Whether the input was an image or a segmentation result
* Different combinations of flags (-check_only, --seg_input, --verbose, -o)

#### 1. Only checking an already conformed image
Arguments: -i, --check_only 
Input: a conformed image
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/T1.mgz --check_only
```
Output:
```
Reading input: /data_dir/T1.mgz ...
Input /data_dir/T1.mgz is already conformed! Exiting.
```

#### 2. Only checking a non-conformed image
Arguments: -i, --check_only 
Input: a non-conformed image
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/001.mgz --check_only
```
Output:
```
Reading input: /data_dir/001.mgz ...
The input image is not conformed.
check_only flag provided. Exiting without conforming input image.
```

#### 3. Only checking a non-conformed image, with verbose output
Arguments: -i, --check_only, --verbose 
Input: a non-conformed image
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/001.mgz --check_only --verbose
```
Output:
```
Reading input: /data_dir/001.mgz ...
The input image is not conformed.
A conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        False
 - Voxel size 1x1x1:              False
 - Orientation LIA:               False
 - Dtype uint8:                   False
check_only flag provided. Exiting without conforming input image.
```

#### 4. Conforming a non-conformed image
Arguments: -i, -o 
Input: a non-conformed image
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/001.mgz -o /data_dir/001_conformed.mgz
```
Output:
```
Reading input: /data_dir/001.mgz ...
The input image is not conformed.
Input:    min: 0.0  max: 2221.3564453125
rescale:  min: 0.0  max: 1090.6860146484375  scale: 0.23379780851246593
Output:   min: 0.0  max: 255.0
Writing conformed image: /data_dir/001_conformed.mgz
```

#### 5. Conforming a non-conformed image, including check_only flag
Arguments: -i, -o , --check_only
Input: a non-conformed image
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/001.mgz -o /data_dir/001_conformed.mgz –check_only
```
Output:
```
ERROR: You passed in check_only. Please do not also specify output image
```
NOTE: As desired, passing in an output filename along with `check_only` causes an error to be raised.

#### 6. Only checking a conformed segmentation
Arguments: -i, --check_only
Input: a conformed segmentation
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/aparc.DKTatlas+aseg.deep.mgz --check_only
```
Output:
```
The input image is not conformed.
check_only flag provided. Exiting without conforming input image.
```
NOTE: This conformed segmentation input was incorrectly classified as not conformed, because the `--seg_input` flag was not used to clarify that the input is a segmentation. Because of this, the `dtype` criteria (which is not necessary for segmentations) is not ignored, and thus is checked and leads to flagging this as not conformed.

#### 7. Only checking a conformed segmentation, including seg_input flag
Arguments: -i, --check_only, --seg_input
Input: a conformed segmentation
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/aparc.DKTatlas+aseg.deep.mgz --check_only --seg_input
```
Output:
```
Input /data_dir/aparc.DKTatlas+aseg.deep.mgz is already conformed! Exiting.
```
NOTE: The previous mistake is avoided by specifying `--seg_input`.

#### 8. Only checking a non-conformed segmentation, including seg_input flag
Arguments: -i, --check_only, --seg_input
Input: a non-conformed segmentation
Command:
```
python3 FastSurferCNN/data_loader/conform.py -i /data_dir/aparc.DKTatlas+aseg.deep.faulty.mgz --check_only --seg_input
```
Output:
```
The input image is not conformed.
check_only flag provided. Exiting without conforming input image.
```
NOTE: The previous mistake is avoided by specifying `--seg_input`.
</details>

AhmedFaisal95:
I added an argument `conformed_img_filename` to the `fastsurfercnn` function.
This is needed particularly when processing multiple subjects, where the path to each conformed image to be saved will differ.

AhmedFaisal95:
Updated with latest from master and integrated `save_image` from #86.
Tested on mgz, nii, and nii.gz input images.

Among the earlier comments, item 6 under eval.py mentions saving the conformed image in a format matching the output segmentation, but we also currently infer the format from the `conformed_name` argument (for e.g. orig.mgz).
Which criteria should decide the output format?
@LeHenschel

recon_surf/recon-surf.sh
```diff
@@ -52,8 +52,9 @@ function usage()
     echo "./recon-surf.sh"
     echo -e "\t--sid <subjectID>             Subject ID for directory inside \$SUBJECTS_DIR to be created"
     echo -e "\t--sd  <subjects_dir>          Output directory \$SUBJECTS_DIR (pass via environment or here)"
-    echo -e "\t--t1  <T1_input>              T1 full head input (not bias corrected)"
-    echo -e "\t--seg <segmentation_input>    Name of intermediate DL-based segmentation file (similar to aparc+aseg). Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz."
+    echo -e "\t--t1  <T1_input>              T1 full head input (not bias corrected). This must be a conformed image (dimensions: 256x256x256, voxel size: 1x1x1,
+    LIA orientation, and data type UCHAR). Images can be conformed using FastSurferCNN's conform.py script."
```
LeHenschel, line 56: I think it would be good to have the command here (python3 conform.py --input <non-conformed-T1-input> --output <T1-input>)

recon_surf/recon-surf.sh
```diff
@@ -52,8 +52,9 @@ function usage()
...
+    echo -e "\t--seg <segmentation_input>    Name of intermediate DL-based segmentation file (similar to aparc+aseg). This must be conformed (dimensions: 256x256x256, voxel size: 1x1x1, LIA orientation). FastSurferCNN's segmentations are conformed by default; please ensure that segmentations produced otherwise are conformed. Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz."
```
LeHenschel, line 57: I think it would be good to have the command here (python3 conform.py --input <non-conformed-segmentation-input> --output <segmentation-input> --seg_input)

FastSurferCNN/README.md
```diff
@@ -13,6 +13,7 @@ The *FastSurferCNN* directory contains all the source code and modules needed to
 * --i_dir: Path to the input volume directory (e.g /your/path/to/ADNI/fs60) or 
 * --csv_file: Path to csv-file listing input volume directories
 * --in_name: name of the MRI_volume (like mri_volume.mgz, __default: orig.mgz__)
+* --conformed_name: name of the conformed MRI_volume (the input volume is always first conformed, if not already, and the result is saved in this file, __default: orig.mgz__)
```
LeHenschel, line 16: maybe "under the given name" instead of "in this file"?

FastSurferCNN/data_loader/conform.py
```diff
@@ -57,9 +57,20 @@ def options_parse():
     parser.add_option('--input', '-i', dest='input', help=h_input)
     parser.add_option('--output', '-o', dest='output', help=h_output)
     parser.add_option('--order', dest='order', help=h_order, type="int", default=1)
+    parser.add_option('--check_only', dest='check_only', default=False, action='store_true',
+                      help='If True, only checks if the input image is conformed')
```
LeHenschel, line 61: specify that no output is returned?

FastSurferCNN/eval.py
```diff
@@ -88,6 +88,8 @@ def options_parse():
                         help='name under which segmentation will be saved. Default: aparc.DKTatlas+aseg.deep.mgz. '
                              'If a separate subfolder is desired (e.g. FS conform, add it to the name: '
                              'mri/aparc.DKTatlas+aseg.deep.mgz)')
+    parser.add_argument('--conf_name', '--conformed_name', dest='conformed_name', default='orig.mgz',
```
LeHenschel, line 91: We also save the original in orig/001.mgz, right? Maybe mention this here as well

recon_surf/README.md
```diff
@@ -20,10 +20,10 @@ List them by running the following command:
 ### Required arguments
 * --sd: Output directory \$SUBJECTS_DIR (equivalent to FreeSurfer setup --> $SUBJECTS_DIR/sid/mri; $SUBJECTS_DIR/sid/surf ... will be created).
 * --sid: Subject ID for directory inside \$SUBJECTS_DIR to be created ($SUBJECTS_DIR/sid/...)
-* --t1: T1 full head input (not bias corrected). 
+* --t1: T1 full head input (not bias corrected). This must be conformed (dimensions: 256x256x256, voxel size: 1x1x1, LIA orientation, and data type UCHAR). Images can be conformed using FastSurferCNN's [conform.py](https://github.com/Deep-MI/FastSurfer/blob/master/FastSurferCNN/data_loader/conform.py) script.
```
LeHenschel, line 23: Add command to do this (python3 conform.py --input <unconformed-T1> --outptu <conformed-T1>)

recon_surf/README.md
```diff
@@ -20,10 +20,10 @@ List them by running the following command:
...
 
 ### Optional arguments
-* --seg: Global path with filename of segmentation (where and under which name to find it, must already exist). Default location: $SUBJECTS_DIR/$sid/mri/aparc.DKTatlas+aseg.deep.mgz
+* --seg: Global path with filename of segmentation (where and under which name to find it, must already exist). This must be conformed (dimensions: 256x256x256, voxel size: 1x1x1, and LIA orientation). FastSurferCNN's segmentations are conformed by default. Please ensure that segmentations produced otherwise are also conformed. Default location: $SUBJECTS_DIR/$sid/mri/aparc.DKTatlas+aseg.deep.mgz
```
LeHenschel, line 26: Also point to conform.py and default command (like it is done in recon-surf.sh argument description)

run_fastsurfer.sh
```diff
@@ -68,6 +69,7 @@ function usage()
     echo -e "\t--sd  <subjects_dir>                   Output directory \$SUBJECTS_DIR (pass via environment or here)"
     echo -e "\t--t1  <T1_input>                       T1 full head input (not bias corrected)"
     echo -e "\t--seg <segmentation_input>             Name of intermediate DL-based segmentation file (similar to aparc+aseg). Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz."
```
LeHenschel, line 71: Maybe we should add a sentence, that this inference is based on the conformed image and hence, the segmentation is already conformed

LeHenschel (commented):
See comments

recon_surf/recon-surf.sh
AhmedFaisal95: In this case, do you mean for `--seg`? If so, I think conform.py currently doesn't support conforming segmentations?

recon_surf/README.md
AhmedFaisal95: Same as the similar comment: I think conform.py currently doesn't support conforming segmentations.

AhmedFaisal95:
> I reviewed the changes and added some suggestions. For the question regarding the saving format of the original image: if the user specifies the conformed_name, this should be the criteria to decide the output format on. We might want to a.) include a warning, that the format of the segmentation and conformed input are not identical + how to change conformed input output type (--conformed_name flag) b.) save by default a version of the input image, which matches the segmentation (in addition to the conformed_name).
> 
> I tend toward option a.). What are your opinions @m-reuter, @AhmedFaisal95?

I also think option a may be better.

FastSurferCNN/eval.py
```diff
@@ -89,7 +89,7 @@ def options_parse():
                              'If a separate subfolder is desired (e.g. FS conform, add it to the name: '
                              'mri/aparc.DKTatlas+aseg.deep.mgz)')
     parser.add_argument('--conf_name', '--conformed_name', dest='conformed_name', default='orig.mgz',
-                        help='Name under which the conformed input image will be saved, in the same directory as the segmentation (the input image is always conformed first, if it is not already conformed). Default: orig.mgz.')
+                        help='Name under which the conformed input image will be saved, in the same directory as the segmentation (the input image is always conformed first, if it is not already conformed). In addition, the original image is always saved in orig/001.mgz in the output directory. Default: orig.mgz.')
```
LeHenschel, line 92: "the original input image is saved in the output directory as $id/mri/orig/001.mgz"

run_fastsurfer.sh
```diff
@@ -68,7 +68,7 @@ function usage()
     echo -e "\t--sid <subjectID>                      Subject ID for directory inside \$SUBJECTS_DIR to be created"
     echo -e "\t--sd  <subjects_dir>                   Output directory \$SUBJECTS_DIR (pass via environment or here)"
     echo -e "\t--t1  <T1_input>                       T1 full head input (not bias corrected)"
-    echo -e "\t--seg <segmentation_input>             Name of intermediate DL-based segmentation file (similar to aparc+aseg). Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz."
+    echo -e "\t--seg <segmentation_input>             Name of intermediate DL-based segmentation file (similar to aparc+aseg). This segmentation is already conformed, since inference is always based on a conformed image. Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz."
```
LeHenschel, line 71: When using FastSurfer, this segmentation is already conformed...

run_fastsurfer.sh
```diff
@@ -273,6 +273,11 @@ if [ -z "$seg" ]
     seg="${sd}/${subject}/mri/aparc.DKTatlas+aseg.deep.mgz"
 fi
 
+if [ -z "$conformed_input" ]
+  then
+    conformed_input="${sd}/${subject}/mri/conformed_orig.mgz"
```
LeHenschel: This should be called orig.mgz

Issue #66, opened 2021-09-14
Running with FreeSurfer v7


LucasCampos:
Hello,

I see in the [`recon_surf.sh` pipeline](https://github.com/Deep-MI/FastSurfer/blob/8fafb9b82e4aa966e7de89eed2d169bd60e5bdfc/recon_surf/recon-surf.sh#L279) that using FastSurfer with FreeSurfer v7 is not recommended. Is this still the case? If so, is the issue that FastSurfer was trained exclusively on FreeSurfer 6.0 data (as per issue #47) or is there something else?

What could be done to run FastSurfer with FreeSurfer v7?

Bests,
Lucas

m-reuter:
Hi, yes, we noticed some reduced test-retest reliability values in FreeSurfer V7. Not sure this is still true for 7.2 (we haven't tested). This was tested and shown directly in vanilla FreeSurfer. FastSurfer inherits those problems if we switch to using the FreeSurfer 7.1 binaries for surface processing. 
Not sure why you want to use FreeSurfer 7? If you want to test things you can either:
- try what happens if you source FS7 instead of 6 and run the pipeline (could fail). This scenario would use FS7 binaries, but the pipeline (recon-surf) would be more similar to how things are done in FS6.
- I started to adopt recon-surf to the 7.1 recon-all steps a while ago and can share that script for testing. 

Anyway, short answer, FS 71 is not recommended, and 7.2 is not tested.

BarryLiu97:
Hi, I'm new here and I have a similar question. And may I ask what's the difference between 6 and 7? From what I saw, 7 is a lot faster than 6.

m-reuter:
For differences between FreeSurfer versions consult their webpages, e.g. 
https://surfer.nmr.mgh.harvard.edu/fswiki/ReleaseNotes 
and mailing list. We are currently exploring basing our surface pipeline on FS72 especially for high-res processing.

m-reuter:
With #114 this has been added to the dev branch and is currently undergoing testing.

m-reuter:
Finally we found the cause for the reduced test retest reliability. We now replace FreeSurfer's global alignment on the sphere with our own code (see #164 ). This has been tested in the dev branch. It was the last thing holding us back to upgrade our FreeSurfer dependency to 7.2 and release a new version of FastSurfer.

PR #114, opened 2022-03-11
Update recon-surf to FreeSurfer 7.2

m-reuter:
This pull requests is targeted at release 1.1 of FastSurfer and moves recon-surf to FreeSurfer 7.2. (see also #66 )

Tests will need to be performed for the different flag options in recon-surf. Also note that maybe the script to extract timings needs to be adopted to the new pipeline. 

Furthermore we need the full testing pipeline eventually (once the v1.1. release is ready) to check if accuracy, group differentiation and test-retest reliability remain high.

## Background: 
recon-all in FreeSurfer changed significantly between FS6.0 and FS7* (see https://surfer.nmr.mgh.harvard.edu/fswiki/ReleaseNotes ). In order to perform similar surface processing I needed to modify recon-surf in many places: with respect to our commands (to make them similar to how FreeSurfer does it now), with respect to some new commands passed to recon-all, and with respect to the new order in which recon-all blocks are called now.  This way we can inherit many minor changes directly via recon-all calls and we can finally drop the distribution of the mris_make_surfaces binary. 

## Improvements: 
- NU correct: the first correction in recon-surf would still use N3 (MNI tools) as passing a mask did not work with the new ANTS N4 in FreeSurfer. Therefore I now directly call N4 from python (via SimpleITK) and added code to read and write mgz format into sITK images. With a brain mask, N4 is fast and the following WM normalisation (to around 110) can be done without the talairach , using a ball (r=50) at the  center of the brain mask instead.

### DIFF
--- a/recon_surf/N4_bias_correct.py
+++ b/recon_surf/N4_bias_correct.py
@@ -21,6 +21,9 @@ import optparse
 import numpy as np
 import sys
 import SimpleITK as sitk
+import nibabel as nib
+import image_io as iio
+
 
 HELPTEXT = """
 
@@ -261,7 +264,8 @@ if __name__ == "__main__":
 
     # read image (only nii supported) and convert to float32
     print("\nreading {}".format(options.invol))
-    itkimage = sitk.ReadImage(options.invol, sitk.sitkFloat32)
+    #itkimage = sitk.ReadImage(options.invol, sitk.sitkFloat32)
+    itkimage, header = iio.readITKimage(options.invol, sitk.sitkFloat32, with_header=True)
 
     # if talaraich is passed
     talorig = None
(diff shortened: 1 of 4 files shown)


m-reuter:
Update: NU Correct is implemented directly via simpleITK including WM scaling to 110 (as this can use the mask while the FreeSurfer wrapper cannot). Also recent updates to master for release 1.0 are integrated and resolved. Still needs some testing before merge.

recon_surf/N4_bias_correct.py
```diff
@@ -0,0 +1,304 @@
...
+Original Author: Martin Reuter
+Date: Mar-18-2022
+"""
+
+h_invol    = 'path to input.nii.gz'
+h_outvol   = 'path to corrected.nii.gz'
+h_mask     = 'optional: path to mask.nii.gz'
+h_shrink   = '<int> shrink factor, default: 4'
+h_levels   = '<int> number of fitting levels, default: 4'
+h_numiter  = '<int> number of iterations per level, default: 4'
```
LeHenschel, line 76: For h_numiter the help text specifies the default as 4, but in the parser option below (line 92) the default is set to 50 (probably copy-paste error from h_levels?)

PR #121, opened 2022-03-17
Fix seg names

LeHenschel:
This pull request contains two majors updates:

1. Correct segmentation names in volume files, labels and stats files from aparc+aseg to aparc.DKTatlas+aseg. DL-prediction is based on aparc.DKTatlas+aseg which should be reflected in the recon-surf output files to avoid confusion.

2. Create symbolic links to original FreeSurfer names from mapped volumes and labels which allows smooth transition in subprocessing steps with FreeSurfer (i.e. sub-segmentation of hippocampus/thalamus or TRACULA)


m-reuter:
Sym links should not be created if --fsaparc is used. We need IF blocks.

LeHenschel:
I moved the symlink for aparc.DKTatlas+aseg.mapped.mgz and wmparc.mapped.mgz into if statements. The symlink to the label files were already included in an if block.

PR #131, opened 2022-03-30
Upgrade: Pytorch, Docker image, and other dependencies

af-a:
## Summary

This PR upgrades the FastSurfer docker image and dependencies to use newer OS and library versions.

The primary upgrades are the following:
* `pytorch`: 1.2.0 --> 1.10.0
* `torchvision`: 0.4.0 --> 0.11.1
* `cudatoolkit`: 10.0 --> 11.3
* `python`: 3.6 --> 3.8

In the base docker image:
* `Ubuntu`: 16.04 --> 20.04
* `CUDA`: 10.0 --> 11.1
* `CuDNN`: 7 --> 8

Other dependencies have been adapted to the requirements of these versions. An updated pip `requirements.txt` is also included.

The upgrade of the pytorch version is motivated by results of tests showing reductions in inference time by ~11% on CPU and ~30% on GPU.

## Caveats

* The resulting base FastSurfer docker image increases in size (due to the newer Ubuntu20.04 image and pytorch version), from approx. 11.3GB to 18GB.
* The current `requirements.txt` has been generated for python3.8. Older versions may not support the required library versions.


LeHenschel:
The commit does not include an update of the FreeSurfer version to FS7.2. If I understand it correctly, this should be included in the same release as the update to FS7.2, right (@m-reuter)? In that case the Dockerfile, Dockerfile_CPU and Dockerfile_reconsurf need to be changed by changing the wget command (line 31):

`wget -qO- https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/7.2.0/freesurfer-linux-centos6_x86_64-7.2.0.tar.gz | tar zxv --no-same-owner -C /opt \`

and the Perl5lib environment variables (lines 75 and 76):
`PERL5LIB=/opt/freesurfer/mni/share/perl5 \
    MNI_PERL5LIB=/opt/freesurfer/mni/share/perl5 \`

af-a:
> The commit does not include an update of the FreeSurfer version to FS7.2. If I understand it correctly, this should be included in the same release as the update to FS7.2, right (@m-reuter)? In that case the Dockerfile, Dockerfile_CPU and Dockerfile_reconsurf need to be changed by changing the wget command (line 31):
> 
> `wget -qO- https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/7.2.0/freesurfer-linux-centos6_x86_64-7.2.0.tar.gz | tar zxv --no-same-owner -C /opt \`
> 
> and the Perl5lib environment variables (lines 75 and 76): `PERL5LIB=/opt/freesurfer/mni/share/perl5 \ MNI_PERL5LIB=/opt/freesurfer/mni/share/perl5 \`

This branch is directly based on the current `dev` branch, which does not have the update to FS72 yet.
I assumed that update would be made in a separate PR, and then the changes here would be adapted following the merging of that.

If not, we could introduce that update directly in this PR.

LeHenschel:
> This branch is directly based on the current dev branch, which does not have the update to FS72 yet.

You mean the stable branch? Because the dev branch has the update to FS7.2 already, hence my question.

See here e.g: https://github.com/Deep-MI/FastSurfer/blob/dev/recon_surf/recon-surf.sh#L19

af-a:
No, I meant the `dev` branch, and by FS72 update I meant to the wget command in the Dockerfiles (sorry if that wan't clear).

In `dev`, `recon-surf.sh` has been updated for FS72, but not the Dockerfiles yet; for e.g.: https://github.com/Deep-MI/FastSurfer/blob/33d2a71f0842d833f69a2deb76ae0c8822c2f52b/Docker/Dockerfile_reconsurf#L31.

LeHenschel:
Yeah, the missing update in the Dockerfile is exactly the problematic point. Because the dev branch (to which your Dockerfiles will be pushed) are already switched to FS7.2. The Dockerfile is not. Hence, it is not possible to run anything in the docker currently because the FreeSurfer version is wrong (https://github.com/Deep-MI/FastSurfer/blob/dev/recon_surf/recon-surf.sh#L284; and also some commands changed between FS60 and FS72 and will fail). If you would add this Dockerfile commit to the stable, it would make more sense, because then it is only the update of the other libs (at FS60).

So we have to decide if we a.) want first an update of the libs on fs60 (stable) or b.) directly update libs and fs (dev), in which case you would need to change the lines I pointed out

So because fs7.2 switch is already in, and the current Dockerfile in dev will fail as well, I would vote for changing the FS7.2 version in this commit as well and merge it asap. @m-reuter, what do you think?

LeHenschel:
Oh, and I think simpleITK is missing, right? That should be included as well.

LeHenschel:
I added simpleITK and FreeSurfer 7.2 to the Dockerfile on dev (#132). Once you add these changes as well, we can merge this pull request.

af-a:
I updated this PR with the aforementioned fixes from `dev`.

PR #134, opened 2022-03-31
Add hires support to conforming and recon-surf

m-reuter:
This PR adds support for processing sub mm images with the recon-surf pipeline using freesurfer 7.2. 

Also the conform check was adopted to allow checking  and performing conform min.


m-reuter:
addresses #146 partially

Issue #138, opened 2022-04-05
Docker Images too large


m-reuter:
## Feature Description
Shrink size of Docker images significantly

## Motivation
FreeSurfer is a beast with more than 10GB, plus dependencies. But also the nvidia docker keeps growing and growing. Why? We don't know but you can guess what happens when we put both together. Images that are larger than 20GB and that is outrageously inappropriate. 

## Alternatives
We can drop quite a few install packages from FreeSurfer (e.g. because we don't want to run freeview or other GUI stuff and we don't want to build FreeSurfer) that will shave off 2GB, then we need to only install what is needed in our pipeline. This is really tricky as FreeSurfer tends to read in files that are not passed on the command line. 

Simultaneously we need to explore alternatives to the nvidia docker images and directly install only the necessary packages and python dependencies. Maybe with a two-stage docker build.

m-reuter:
This has been addressed with various pull requests: #139  #141  #145
significantly reducing required size of docker images.

Issue #33, opened 2020-09-18
[Question] Space of orig and aparc.DKTatlas+aseg


tashrifbillah:
The training [input](https://github.com/Deep-MI/FastSurfer/blob/master/FastSurferCNN/generate_hdf5.py#L172) for FastSurfer is `mrig/orig.mgz` and `mri/aparc.DKTatlas+aseg.mgz`. My question--are the two files in any standard space such as MNI? I know the two are created by FreeSurfer but I do not know the answer.

I noticed at the top of FreeSurfer, `mri_convert -c` is applied that makes `orig.mgz` with the following:

```vim
voxel to ras transform:
               -1.0000   0.0000   0.0000   128.5000
                0.0000   0.0000   1.0000  -127.5000
                0.0000  -1.0000   0.0000   128.5000
                0.0000   0.0000   0.0000     1.0000
```

m-reuter:
No this is still in subject space, the only thing FS does is conform it to 1mm isotropic and rotate it to a standard position (according to the matrix that you show). This ensures that the correct directions are along the right axes. Further, it robustly maps intensites to UCHAR (0..255). 
If your inputs are not conform, FastSurfer should make them conform (we have a python script that basically mimics the mri_convert conform command).

tashrifbillah:
Hi @m-reuter ,

I understand the conform step is materialized via `scipy.ndimage.affine_transform`:
https://github.com/Deep-MI/FastSurfer/blob/196d9f7c021cc7961264f6cfa9ea31f4e2423cdd/FastSurferCNN/data_loader/conform.py#L66

```python
new_data = affine_transform(img.get_data(), inv(vox2vox), output_shape=out_shape, order=order)
```


cc @sbouix.

m-reuter:
Yes, the voxel to voxel transform can have scaling if the voxel sizes are not already 1mm isotropic.

PR #139, opened 2022-04-05
Feature: shrink docker - FS dependencies

m-reuter:
This PR removes several dependencies from the Docker images with FreeSurfer, shaving away around 2GB of data. 

It has not yet been fully tested for all parameter settings, but worked for default. 

This is addressing #138

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -7,45 +7,46 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Install custom libraries + freesurfer 6.0 (and necessary dependencies)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-         build-essential \
-         cmake \
-         git \
-         vim \
-         wget \
-         ca-certificates \
-         bzip2 \
-         libx11-6 \
-         libjpeg-dev \
-         libpng-dev \
-         bc \
-         tar \
-         zip \
-         gawk \
-         tcsh \
-         time \
-         libgomp1 \
-         libglu1-mesa \
-	 libglu1-mesa-dev \
-	 perl-modules && \
-         apt-get clean && \
-         rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
-	 wget -qO- https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/7.2.0/freesurfer-linux-centos6_x86_64-7.2.0.tar.gz | tar zxv --no-same-owner -C /opt \
-    	--exclude='freesurfer/trctrain' \
-    	--exclude='freesurfer/subjects/fsaverage_sym' \
-    	--exclude='freesurfer/subjects/fsaverage3' \
-    	--exclude='freesurfer/subjects/fsaverage4' \
-    	--exclude='freesurfer/subjects/fsaverage5' \
-    	--exclude='freesurfer/subjects/fsaverage6' \
-    	--exclude='freesurfer/subjects/cvs_avg35' \
-    	--exclude='freesurfer/subjects/cvs_avg35_inMNI152' \
-    	--exclude='freesurfer/subjects/bert' \
-    	--exclude='freesurfer/subjects/V1_average' \
-    	--exclude='freesurfer/average/mult-comp-cor' \
-    	--exclude='freesurfer/lib/cuda' \
-    	--exclude='freesurfer/lib/qt'
+      wget \
+      git \
+      tcsh \
+      time \
+      bc \
+      gawk \
+      libgomp1 && \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    wget --no-check-certificate -qO- https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/7.2.0/freesurfer-linux-ubuntu18_amd64-7.2.0.tar.gz  | tar zxv --no-same-owner -C /opt \
+      --exclude='freesurfer/trctrain' \
+      --exclude='freesurfer/subjects/fsaverage_sym' \
+      --exclude='freesurfer/subjects/fsaverage3' \
+      --exclude='freesurfer/subjects/fsaverage4' \
+      --exclude='freesurfer/subjects/fsaverage5' \
+      --exclude='freesurfer/subjects/fsaverage6' \
+      --exclude='freesurfer/subjects/cvs_avg35' \
+      --exclude='freesurfer/subjects/cvs_avg35_inMNI152' \
+      --exclude='freesurfer/subjects/bert' \
+      --exclude='freesurfer/subjects/V1_average' \
+      --exclude='freesurfer/average/mult-comp-cor' \
+      --exclude='freesurfer/lib/cuda' \
+      --exclude='freesurfer/lib/qt' \
+      --exclude='freesurfer/matlab' \
+      --exclude='freesurfer/diffusion' \
+      --exclude='freesurfer/bin/freeview.bin' \
+      --exclude='freesurfer/bin/freeview' \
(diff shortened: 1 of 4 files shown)


af-a:
I can't seem to build the recon_surf docker image (from Dockerfile_reconsurf) in its current state; I get the error:
```
Step 8/12 : RUN python$PYTHON_VERSION -m pip install -U git+https://github.com/Deep-MI/LaPy.git#egg=lapy &&     python$PYTHON_VERSION -m pip install simpleitk==2.1.1
 ---> Running in 7a64b0e1724e
Collecting lapy
  Cloning https://github.com/Deep-MI/LaPy.git to /tmp/pip-install-way_kyrs/lapy_37aa311933e94d5c914410d5e490bcc1
  Running command git clone -q https://github.com/Deep-MI/LaPy.git /tmp/pip-install-way_kyrs/lapy_37aa311933e94d5c914410d5e490bcc1
  fatal: unable to access 'https://github.com/Deep-MI/LaPy.git/': server certificate verification failed. CAfile: none CRLfile: none
WARNING: Discarding git+https://github.com/Deep-MI/LaPy.git#egg=lapy. Command errored out with exit status 128: git clone -q https://github.com/Deep-MI/LaPy.git /tmp/pip-install-way_kyrs/lapy_37aa311933e94d5c914410d5e490bcc1 Check the logs for full command output.
ERROR: Could not find a version that satisfies the requirement lapy (unavailable) (from versions: none)
ERROR: No matching distribution found for lapy (unavailable)
The command '/bin/sh -c python$PYTHON_VERSION -m pip install -U git+https://github.com/Deep-MI/LaPy.git#egg=lapy &&     python$PYTHON_VERSION -m pip install simpleitk==2.1.1' returned a non-zero code: 1
```
I think the issue is that `LaPy` is not found: `fatal: unable to access 'https://github.com/Deep-MI/LaPy.git/': server certificate verification failed.`

Tested this on two machines, with the same results:
1. Ubuntu 20.04.4 LTS (`Linux ... 5.13.0-30-generic #33~20.04.1-Ubuntu SMP Mon Feb 7 14:25:10 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`)
2. CentOS Linux 7 (Core) (`Linux ... 3.10.0-1160.49.1.el7.x86_64 #1 SMP Tue Nov 30 15:51:32 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`)

I was able to fix this by including the `ca-certificates` package, which was recently removed from the Dockerfile.
I think we may need to add that back in.

Docker/Dockerfile
```diff
@@ -7,45 +7,46 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Install custom libraries + freesurfer 6.0 (and necessary dependencies)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-         build-essential \
-         cmake \
-         git \
-         vim \
-         wget \
-         ca-certificates \
```
af-a, line 15: We may need to include `ca-certificates` again, due to the aforementioned build error.

Docker/Dockerfile_CPU
```diff
@@ -7,61 +7,48 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Install custom libraries + freesurfer 6.0 (and necessary dependencies)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-         build-essential \
-         cmake \
-         git \
-         vim \
-         wget \
-         ca-certificates \
```
af-a, line 15: Same

Docker/Dockerfile_reconsurf
```diff
@@ -6,67 +6,54 @@ ARG PYTHON_VERSION=3.8
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Install custom libraries + freesurfer 6.0 (and necessary dependencies)
-RUN apt-get update && apt-get install -y --no-install-recommends \
-         build-essential \
-         cmake \
-         git \
-         vim \
-         wget \
-         ca-certificates \
```
af-a, line 15: Same

af-a:
`recon-surf.sh` also seems to fail at the following recon-all call:
https://github.com/Deep-MI/FastSurfer/blob/c30362d5bfb219f93ccc706e5bafeb0bb221c556/recon_surf/recon-surf.sh#L528

I observe the error:
```
#--------------------------------------------                                   
#@# Fill Wed Apr  6 09:10:42 UTC 2022                                           
FREESURFER: Undefined variable.                                                 
@#@FSTIME  2022:04:06:09:06:13 recon-all N 7 e 268.75 S 3.49 U 264.61 P 99% M 1086920 F 0 R 1751580 W 0 c 493 w 931 I 0 O 10896 L 3.44 1.67 0.68
@#@FSLOADPOST 2022:04:06:09:10:42 recon-all N 7 3.14 2.83 1.4
```

This occurs in different flag combinations (--fsaparc --fstess, --fsaparc, --fstess, and the defaults).

LeHenschel:
Yes, this is an error in FS7.2. At one point instead of FREESURFER_HOME they set the variable to FREESURFER inside the recon-all fill command (https://github.com/freesurfer/freesurfer/blob/fs-7.2/scripts/recon-all#L3028). We can either call mri_fill directly (`mri_fill -a ../scripts/ponscc.cut.log -segmentation aseg.presurf.mgz -ctab /opt/freesurfer/SubCorticalMassLUT.txt -auto-man filled.auto.mgz filled.mgz ../tmp/filled.edits.txt wm.mgz filled.mgz` ) in which case the error does not occur or just export FREESURFER=FREESURFER_HOME (set this in inside docker or during the call).

It is also a bit odd, that we get to the auto-man part. FS_ALLOW_FILLED_EDIT is turned on by default for some reason (https://github.com/freesurfer/freesurfer/blob/fs-7.2/scripts/recon-all#L454). This variable did not exist in FS6.0. Here, the mri_fill command was: `mri_fill -a ../scripts/ponscc.cut.log -xform transforms/talairach.lta -segmentation aseg.auto_noCCseg.mgz wm.mgz filled.mgz`. The filled.auto.mgz exists, but not sure where it is generated. This should only be for manual edits to filled.mgz

m-reuter:
I'll update the PR with the ca-certificates. 

We should export the FREESURFER variable in recon-surf so that it is automatically done independent of the container or direct install. Can one of you do that? Also is this a know bug in FreeSurfer or do we need to report it?

I also noticed in some log output that label2vol complained that no --seed is given (probably it is called from inside another script and not directly from recon-all). This is probably also a bug in FS7.2. Take a look at the recon-surf log to see if you find it and then we need to check if we have to report (or fix) it in FreeSurfer.

m-reuter:
I'll also add the FREESURFER export to this pull request.

af-a:
> I also noticed in some log output that label2vol complained that no --seed is given (probably it is called from inside another script and not directly from recon-all). This is probably also a bug in FS7.2. Take a look at the recon-surf log to see if you find it and then we need to check if we have to report (or fix) it in FreeSurfer.

I do consistently see this message in the `recon-surf.log`s:
```
mri_label2vol supposed to be reproducible but seed not set
```

LeHenschel:
The FREESURFER bug has been fixed in the dev version of FreeSurfer (so it should be known).

m-reuter:
> I do consistently see this message in the `recon-surf.log`s:
> 
> ```
> mri_label2vol supposed to be reproducible but seed not set
> ```

Yes, that is what I mean. But 
- where exactly is this called from (maybe multiple places and I think not from recon-all but somewhere deeper)
- does that script allow passing a --seed and if so, why is it not passing the seed deeper to label2vol
- is this already a known bug and fixed in freesurfer dev or not

I would not worry about it too much now, but if it is unknown to them we should at least report it.

PR #153, opened 2022-04-27
Fix: Include freesurfer mris_topo_fixer command

af-a:
## Description

With freesurfer 7.2 (the default for FastSurfer v1.1), `mris_fix_topology` falls back on the `mris_topo_fixer` command if it fails. This fallback command was removed from the install, so any runs in which `mris_fix_topology` fails would faild completely.

This PR includes the command in the freesurfer install.


m-reuter:
Info: this needs to be included as recon-all falls back to this binary in some cases.

PR #159, opened 2022-05-24
get rca-config from freesurfer dev

m-reuter:
In the meantime freesurfer dev has a fixed rca-config that additionally fixes problems with -s flag

see:

- https://github.com/freesurfer/freesurfer/commit/c10e63484a3cfda5c272641241427fa154777cc2 
- https://github.com/freesurfer/freesurfer/issues/972

so instead of our "sed" fix, we now download this file from dev.


m-reuter:
testing was successful, so we can include these upstream fixes into dev now.

PR #168, opened 2022-07-01
[WIP] VINN Update

LeHenschel:
This branch updates the network to the newest VINN version (Henschel L, Kügler D, Reuter M. FastSurferVINN: Building resolution-independence into deep learning segmentation methods-A solution for HighRes brain MRI. Neuroimage. 2022 May 1;251:118933. doi: 10.1016/j.neuroimage.2022.118933. Epub 2022 Feb 3. PMID: 35122967.). Running the old CNN model is still possible as well. The commits include significant updates to the code base, including loading/saving of config files, new classes for integration with tensorboard, logging and optimizer/lr-scheduler selection. 

Recently tested for running predictions on HCP-TRT as well as training on the original multi-resolution dataset from the publication.

Still missing are updates to the README (run commands and usage explanation), checkpoints, corresponding Docker files, containers, requirements.txt as well as Tutorials. 

Also still required is integration into recon-surf with the updated hires pipeline and final testing (TRT and Group Analysis)


dkuegler:
@AhmedFaisal95 / @LeHenschel 

In the current feature/vinn https://github.com/Deep-MI/FastSurfer/pull/168/commits/03c0af0c7d477d1097b5f7c58c418ed2d6ee8946 If I just pass the sagittal cfg and checkpoint, it crashes

> Traceback (most recent call last):
>   File "FastSurferCNN/run_prediction.py", line 434, in <module>
>     pred_data = eval.get_prediction() #
>   File "FastSurferCNN/run_prediction.py", line 237, in get_prediction
>     pred_prob = self.run_model(pred_prob)
>   File "FastSurferCNN/run_prediction.py", line 165, in run_model
>     pred_prob = self.model.run(self.orig_filename, self.orig_data, self.orig.header.get_zooms(),
>   File "/home/kueglerd/projects/FastSurfer/FastSurferCNN/eval.py", line 172, in run
>     pred_prob = self.eval(test_data_loader, pred_prob, out_scale=out_res)
>   File "/home/kueglerd/projects/FastSurfer/venv/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
>     return func(*args, **kwargs)
>   File "/home/kueglerd/projects/FastSurfer/FastSurferCNN/eval.py", line 154, in eval
>     pred_prob[start_index:start_index + pred.shape[0], :, :, :] += torch.mul(pred, 0.2)
> RuntimeError: The size of tensor a (51) must match the size of tensor b (79) at non-singleton dimension 3

This is due to the output `pred` being initialized with the model's class number (51) but this tries to save it into the view_aggration tensor `pred_prob` which is (79) classes.

af-a:
> @AhmedFaisal95 / @LeHenschel
> 
> In the current feature/vinn [03c0af0](https://github.com/Deep-MI/FastSurfer/commit/03c0af0c7d477d1097b5f7c58c418ed2d6ee8946) If I just pass the sagittal cfg and checkpoint, it crashes
> 
> > Traceback (most recent call last):
> > File "FastSurferCNN/run_prediction.py", line 434, in 
> > pred_data = eval.get_prediction() #
> > File "FastSurferCNN/run_prediction.py", line 237, in get_prediction
> > pred_prob = self.run_model(pred_prob)
> > File "FastSurferCNN/run_prediction.py", line 165, in run_model
> > pred_prob = self.model.run(self.orig_filename, self.orig_data, self.orig.header.get_zooms(),
> > File "/home/kueglerd/projects/FastSurfer/FastSurferCNN/eval.py", line 172, in run
> > pred_prob = self.eval(test_data_loader, pred_prob, out_scale=out_res)
> > File "/home/kueglerd/projects/FastSurfer/venv/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
> > return func(*args, **kwargs)
> > File "/home/kueglerd/projects/FastSurfer/FastSurferCNN/eval.py", line 154, in eval
> > pred_prob[start_index:start_index + pred.shape[0], :, :, :] += torch.mul(pred, 0.2)
> > RuntimeError: The size of tensor a (51) must match the size of tensor b (79) at non-singleton dimension 3
> 
> This is due to the output `pred` being initialized with the model's class number (51) but this tries to save it into the view_aggration tensor `pred_prob` which is (79) classes.

I encountered this in my initial tests (since https://github.com/Deep-MI/FastSurfer/pull/168/commits/60e7c6d71f2cb94484088f10bcbbd35585abc27e) and discussed it. From what I understood,
FastSurfer is not usually run in the single-plane case and so it is not explicitly covered, but @LeHenschel can correct me if I am wrong.

dkuegler:
> I encountered this in my initial tests (since [60e7c6d](https://github.com/Deep-MI/FastSurfer/commit/60e7c6d71f2cb94484088f10bcbbd35585abc27e)) and discussed it. From what I understood, FastSurfer is not usually run in the single-plane case and so it is not explicitly covered, but @LeHenschel can correct me if I am wrong.

that makes sense, but I am trying to hunt down why my predictions with the memory optimizations are different and I basically have narrowed it down to the view aggregration so I need single-plane prediction to detect the bugs.

af-a:
> > I encountered this in my initial tests (since [60e7c6d](https://github.com/Deep-MI/FastSurfer/commit/60e7c6d71f2cb94484088f10bcbbd35585abc27e)) and discussed it. From what I understood, FastSurfer is not usually run in the single-plane case and so it is not explicitly covered, but @LeHenschel can correct me if I am wrong.
> 
> that makes sense, but I am trying to hunt down why my predictions with the memory optimizations are different and I basically have narrowed it down to the view aggregration so I need single-plane prediction to detect the bugs.

I see. I think either coronal or axial on their own would work.

dkuegler:
https://github.com/Deep-MI/FastSurfer/pull/177 a pull request to pull into a pull request :-)

FastSurferCNN/run_prediction.py
```diff
@@ -0,0 +1,419 @@
...
+        if args.run_viewagg_on == "gpu":
+            # run view agg on the gpu (force)
+            self.small_gpu = False
+
+        elif args.run_viewagg_on == "check":
+            # check, if GPU is big enough to run view agg on it
+            # (this currently takes only the total memory into account, not the occupied on)
+            total_gpu_memory = sum([torch.cuda.get_device_properties(i).__getattribute__("total_memory") for i in
+                                    range(torch.cuda.device_count())])
+            self.small_gpu = total_gpu_memory < 8000000000
```
dkuegler: this number needs to be adjusted for VINN, it now depends on the size of the image:
formula for memory required for view aggregation:
`width (up to 320) * height (up to 320) * depth (up to 320) * classes (79) * sizeof(float) (4)` which is about 9.64 GB

FastSurferCNN/run_prediction.py
```diff
@@ -379,7 +379,10 @@ def check_and_download_ckpts(ckptpath):
                         default=os.path.join(os.path.dirname(__file__),
                                              "config/FastSurferVINN_sagittal.yaml"), type=str)
 
-    parser.add_argument('--no_cuda', action='store_true', default=False, help="Disables GPU usage")
+    parser.add_argument('--no_cuda', help="Deprecated: Use --device flag instead!")
+    parser.add_argument('--device', default="cuda",
```
dkuegler: There should probably an "auto" option for device, which should be the default.

This is so the default configuration works independent of the hardware of the PC.

I think we also had some automatic identification of Hardware in fastsurfer before.

dkuegler: I would even go as far as raising an exception, if cuda is passed as the device, but torch does not run on the GPU (e.g. the device is invalid)

dkuegler: Similarly, batch_size should probably also default to some automatic behavior if no number is given to ensure the default configuration most likely works.

I am going to create a dedicated issue for this, since this is not really required for v2.0 and can be done in a later update.

LeHenschel: Currently, we just switch to cpu if gpu is set, but not available and return a warning. Assertion works as well. For auto, this only makes sense in my opinion once we include the other options (MAC M3 and what not). @m-reuter wanted to create a check list for that.

dkuegler: My idea is that if you specifically specify cuda, you should get an error if it can't do it on the GPU.

This is a clear expected behavior that also lets you debug well. 

See also https://peps.python.org/pep-0020/

LeHenschel: Yeah, I am not arguing against it. I would just change it in one go with the auto option.

m-reuter:
I just merged and updated the default checkpoints download function and command line script. And I updated the dockerfiles to download the checkpoints and to point to the correct entry script for the Segmentation only ones: run_prediction. We need to consider always using run_fastsurfer as entry point to avoid these kinds of breaking changes when network scripts get changed.

dkuegler:
should we also prune the FastSurferCNN checkpoints out of the git tree with the 2.0 release then?

FastSurferCNN/run_prediction.py
m-reuter: Reply I updated it to "auto" so it will be easier in the future to add more options.

m-reuter:
> should we also prune the FastSurferCNN checkpoints out of the git tree with the 2.0 release then?

Probably not, as this will break things if someone checks out an older release tag. Maybe in the distant future. But they can be deleted and downloaded from now on.

FastSurferCNN/eval.py
```diff
@@ -109,8 +110,18 @@ def set_model(self, cfg=None):
...
+        # workaround for mps (directly loading to map_location=mps results in zeros)
+        if (self.device.type == 'mps'):
+            model_state=torch.load(ckpt, map_location='cpu')
+            self.model.to('cpu')
+            self.model.load_state_dict(model_state['model_state'])
+            self.model.to(self.device)
+        else:
+            model_state = torch.load(ckpt, map_location=self.device)
+            self.model.load_state_dict(model_state['model_state'])
+        #print(model_state['model_state']['inp_block.bn0.weight'])
```
LeHenschel: Remove the print statements?

LeHenschel (commented):
Shouldn't we rather change the download script in this case? VINN should be the default and defining all and vinn is a bit counter intuitive

FastSurferCNN/download_checkpoints.py
```diff
@@ -0,0 +1,50 @@
...
+
+import argparse
+import os
+
+from utils.checkpoint import check_and_download_ckpts, get_checkpoints, VINN_AXI, VINN_COR, VINN_SAG, URL
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Check and Download Network Checkpoints')
+    parser.add_argument("--all", default=False, action="store_true",
```
LeHenschel, line 25: I would rather make this a flag with which you can provide different arguments (all, vinn, cnn) and use all as the default

dkuegler: Currently, vinn should be the only and automatic option. If people want to use faalstsurfercnn, they should use fastsurfer 1.X imo
If we integrate other modules for other task, it should be a list of modules (e.g. aseg, aparc)

FastSurferCNN/eval.py
m-reuter: Ok, can you do this.

FastSurferCNN/download_checkpoints.py
m-reuter: I want an all flag to download everything that is available. This will be used in docker. 
Integration of fscnn well be decided based on vinn performance in tests.

LeHenschel: My main issue here is that there should not be multiple flags (--all --vinn --cnn). Especially because at the  moment you need to specify both (--all and --vinn). "all" should already include "vinn" at the very least. Otherwise it is very counterintuitive.

...and faalstsurfercnn is pretty hilarious (=failiest in dutch).

m-reuter: all does include vinn, see the fixes below. I agree that we need to think about having a --flag for each module later or having a --module string or list. I think there is really no difference here. What is more important is to think about good names for the modules. vinn is not great as many module will use the vinn architecture. Probably should be fsvinn for FastsurferVINN . Names cannot be too long.

LeHenschel: Ah, ok. Did not see the new commit. 

Could also be a combination of network and structure (similar to what david suggested --arch vinn and --module aseg, aparc, olfbulb...). Downside is that probably only certain combinations are possible. I get your point and agree that we need a better description, but FastSurfer (therefore fsvinn and FastSurferVINN) are in my opinion not exclusive to the aparc segmentation (just like you also have hippo/thalamus submodules in FreeSurfer).

m-reuter: another downside of the arch and module is that people would not usually know what arch was used for a task. They just want a specific module and don't care what arch that uses. Which is why we should try to only have one candidate for a specific task. Maybe it should be aparcDKT , cerebellum, olfactorybulb, hypothalamus, choroidplexus, piriformcortex etc.

LeHenschel: Sounds good. Could also just have a default and for people who want to try it out can change it (depending if there actually are different architectures for one submodule).

dkuegler: Naming of modules should be consistent with output segmentation file names (so aseg+aparcDKTatlas)

FastSurferCNN/utils/checkpoint.py
```diff
@@ -142,6 +142,7 @@ def download_checkpoint(download_url, checkpoint_name, checkpoint_path):
     except requests.exceptions.HTTPError as e:
         print('[checkpoint] Response code: {}'.format(e.response.status_code))
```
dkuegler: Should this not be logged so it is also in the log files?

af-a: Yes. We can add a LOGGER instance in `checkpoint.py` to log this message as well.

Issue #108, opened 2022-02-21
RuntimeError : CUDA out of memory


wtmccuddy:
Hello, 

Looking for any guidance on GPU memory issue. I read all of the resolved comments on the issue and tried reducing batch size to 1 and checking output of nvidia-smi to kill any python processes using up GPU memory (didn't see any extra stuff running to kill; see output below). Reducing batch size did not help. It seems to me that my GPU should have enough memory for this, but maybe I'm wrong. I was able to get code to run on CPU (running now and no errors yet). Any assistance or information on how to resolve GPU memory issues would be appreciated. 

Thanks!

PS: Using linux machine running ubuntu 20.04.4

This is the error:
RuntimeError: CUDA out of memory. Tried to allocate 4.94 GiB (GPU 0; 1.82 GiB total capacity; 6.93 MiB already allocated; 286.31 MiB free; 8.00 MiB reserved in total by PyTorch)

This is the command I used to run and the subsequent output:
`(base) :~/Desktop/Sub001$ run_fastsurfer.sh --t1 /home/Desktop/Sub001//output_VBG/sub-JW001/sub-JW001_T1_nat_4parc.mgz --sid sub-JW001 --sd /home/Desktop/Sub001//output_VBG/sub-JW001/sub-JW001fastsurfer --parallel --fsaparc --threads 8 --fs_license /usr/local/freesurfer/7.2.0/license.txt --py python 
Sun 20 Feb 2022 06:25:59 PM MST

/usr/local/KUL_apps/FastSurfer/FastSurferCNN ~/Desktop/Sub001
[...]

Issue #12, opened 2020-07-13
fastsurferCNN eval.py memory usage


BBBBryanPan:
Hi,
I downloaded your code and tried to run eval.py in fastsurferCNN folder with this command
`python eval.py --i_dir 'adni' --o_dir output --in_name 1.nii --order 1 --batch_size 1 --simple_run false` 
our computer has 32GB ram and our GPU is 11GB 2080ti. 
Before running the code, I check the free memory is 28GB. Before running
`_, prediction_image = torch.max(torch.add(torch.mul(torch.add(prediction_probability_axial,
                                                                  prediction_probability_coronal), 0.4),
                                              torch.mul(prediction_probability_sagittal, 0.2)), 3)`
the free memory is 8GB and then the script exit unexpectedly with no warning. 
I also separate the code into 4 sentence. After running 
`a= torch.add(prediction_probability_axial, prediction_probability_coronal)`
the free memory is 2.6GB. And exit dealing with the `b = torch.mul(a,0.4)`. 
I believe it should be a problem with the memory, and it's very similar with Issue #2.

LeHenschel:
Hey,

I will have a look at it, but it is possible that the view aggregation does indeed need a lot of CPU memory due to the large number of feature maps (95 classes).

Best,
Leonie

BBBBryanPan:
Thanks a lot. And may I ask how many memory at least is required to run fastsurferCNN/eval.py for one image with size like (179x256x256)?

m-reuter:
Your RAM should be sufficient (at least if you run step i CNN segmentation on the GPU, where you might need to reduce batch size, e.g. try 8). On our GPU it takes 40 sec with 8GB RAM (RTX 2080). For the part ii, 32 Gig CPU RAM should really be enough.

BBBBryanPan:
OK, thanks, I'll try that. Another thing is that does your network support nifti format input? We use nifti format to store our brain mri images.

LeHenschel:
Hey,

yes, nifti format is supported as well (see also #20 ). 

Best,
Leonie

BBBBryanPan:
Hey,
I just find out a solution to decrease the memory spending.  Since 3 variable：`prediction_probability_axial，prediction_probability_coronal， prediction_probability_sagittal` is no longer used after 
[...]

Issue #10, opened 2020-07-07
Memory usage


KobaCemal:
Hello,

I am trying to run FastSurfer on a single image with this code: 

`./run_fastsurfer.sh --t1 /home/koba/Desktop/sub-001/anat/orig.mgz                     --sid subject1 --sd  /home/koba/Desktop/sub-001/fsout                    --parallel --py python3.7 --batch 2`

The script fills my swap memory after Coronal Testing and stops no matter what. It doesn't return any error codes. If I set the batch size >10, it tells me this: 

`RuntimeError: CUDA out of memory. Tried to allocate 320.00 MiB (GPU 0; 3.95 GiB total capacity; 2.49 GiB already allocated; 163.31 MiB free; 3.04 GiB reserved in total by PyTorch)`

My computer has 16 gb of ram (2 gb swap) and nvidia geforce 1050 gpu ( 4 gb)

Does that mean my computer is just insufficient for the code? Or Am I missing something in the installation?

Thank you very much!

LeHenschel:
Hey,

yes, I think this is most likely a memory problem with your GPU. I am not sure if you tried to set the batch size to 1? In addition you can check if you can run the network on your cpu (add the --no_cuda flag):

`./run_fastsurfer.sh --t1 /home/koba/Desktop/sub-001/anat/orig.mgz --sid subject1 --sd /home/koba/Desktop/sub-001/fsout --parallel --py python3.7 --batch 1 --no_cuda --seg_only` 

Best,
Leonie

KobaCemal:
Hi,

Thank you for your quick response! 

I ran the command you suggested with 16 batches. It has been running for 20 minutes. It occupies 10 gb of ram and 6 processors and still going on. Since it is "segmentation only", I think there is something wrong.  Maybe the computer is not enough for the procedure.

I was going to check myself but since the analysis doesn't finish, I think it is okay to ask here. Do the results require any manual correction like FreeSurfer's results do? I have read your paper and couldn't see any indication of manual correction for FreeSurfer's or FastSurfer's results. Do you think it is safe to skip that part (except the most obvious errors of course)? 

I am asking because I have a large dataset and detailed manual correction is something that I want to avoid. It wouldn't be possible to make healthy judgements on so many subjects. I would be happy to hear your opinion on this.

Thank you!

m-reuter:
You can also try batch 1 and with the GPU. 
[...]

PR #176, opened 2022-08-02
Fix the memory issues of the VINN branch

dkuegler:
Several optimizations for the VINN eval script, specifically reduces memory consumption of the network!

### DIFF
--- a/FastSurferCNN/run_prediction.py
+++ b/FastSurferCNN/run_prediction.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 # IMPORTS
+from typing import Tuple, Optional, Union, Dict
+
 import numpy as np
 import torch
 import os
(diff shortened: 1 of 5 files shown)


dkuegler:
@AhmedFaisal95 or @LeHenschel Someone please double-check on their machine

Ideally, we also run this and the old version to 1-2 subjects and make sure no differences are there (which I do not really expect)
This is mainly due to the part where I don't do the conversion to 320³ from 256³ any more (i.e. this is optimized out for eval), but I also have not changed if this breaks the training due to some unexpected side-effect.

dkuegler:
I had a couple of broken merge things... should be fixed here

dkuegler:
As additional point features, this pull request has better documentation for CUDA out of memory errors #108 , generally reduces the memory footprint by using half-precision during view aggregation #10 , #12

LeHenschel:
Checked the pull request on 80 subjects from the validation cohort (40 1.0 mm, 20 0.8 mm and 20 0.7 mm). Results are consistent with the original version (minor variations due to noise). Only issue is the get_num_classes function is not correctly called, leading to an incorrect shape of the prediction tensor.

FastSurferCNN/run_prediction.py
```diff
@@ -160,71 +169,57 @@ def set_model(self, plane):
...
+    def get_gt(self) -> Tuple[nib.analyze.SpatialImage, np.ndarray]:
         return self.gt, self.gt_data
 
-    def get_orig(self):
+    def get_orig(self) -> Tuple[nib.analyze.SpatialImage, np.ndarray]:
         return self.orig, self.orig_data
 
     def get_prediction(self):
-        oh, ow, od = self.orig.shape
+        shape = self.orig.shape + (self.model.get_num_classes(),)
```
LeHenschel, line 185: this needs to be changed to shape = self.orig.shape + (self.get_num_classes(),)

LeHenschel (approved):

PR #177, opened 2022-08-03
[run_prediction] VERIFIED ON BERT; simplify get_prediction memory all…

dkuegler:
…ocation rules, add get_num_classes, fix single-plane prediction

### DIFF
--- a/FastSurferCNN/run_prediction.py
+++ b/FastSurferCNN/run_prediction.py
@@ -184,56 +184,36 @@ class RunModelOnData:
         if self.view_ops["coronal"]["cfg"] is not None:
             if self.view_ops["axial"]["cfg"] is not None or self.view_ops["sagittal"]["cfg"] is not None:
                 self.set_model("coronal")
-            if not self.small_gpu:
-                pred_prob = torch.zeros((self.dim, self.dim, od, self.model.get_num_classes()),
-                                        dtype=torch.float).to(self.device)
-            else:
-                pred_prob = torch.zeros((self.dim, self.dim, od, self.model.get_num_classes()),
-                                        dtype=torch.float)
+            pred_prob = torch.zeros((self.dim, self.dim, od, self.get_num_classes()),
+                                    dtype=torch.float, device='cpu' if self.small_gpu else self.device)
             pred_prob = self.run_model(pred_prob)
 
         # axial inference
         if self.view_ops["axial"]["cfg"] is not None:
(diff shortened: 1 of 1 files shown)


m-reuter:
usually we wait with merging until someone has reviewed the pull request.

FastSurferCNN/run_prediction.py
```diff
@@ -184,56 +184,36 @@ def get_prediction(self):
...
-                                        dtype=torch.float)
+            pred_prob = torch.zeros((self.dim, self.dim, od, self.get_num_classes()),
+                                    dtype=torch.float, device='cpu' if self.small_gpu else self.device)
             pred_prob = self.run_model(pred_prob)
 
         # axial inference
         if self.view_ops["axial"]["cfg"] is not None:
             LOGGER.info("Run axial view agg")
-            if self.view_ops["coronal"]["cfg"] is not None or self.view_ops["sagittal"]["cfg"] is not None:
+            if self.view_ops["coronal"]["cfg"] is not None:
```
m-reuter, line 194: @dkuegler why has the "or" part been removed here and not in the other views?

dkuegler:
Well, this is a merge into the vinn branch, not main fastsurfer.

FastSurferCNN/run_prediction.py
dkuegler: The condition is that "if you already have allocated the memory" (i.e. pred_prob exists) and it only exists on one of the two "or" clauses.

dkuegler:
For reference, I ran the modifications (1) on the baseline and (2) on this modification and the results are identical. Additionally, we can now do single-view predictions (which previously just crashed)

FastSurferCNN/run_prediction.py
m-reuter: above (l185) and below (l208) it is checking the other two views, here only one. Still don't get why.

LeHenschel: It should check for both because afterwards the model is set to the appropriate view. If you only run one view, this is not necessary, but as soon as one of the other has been run below, the next subject needs to start at coronal again and should therefore be reset to it (specifically the checkpoints).

m-reuter: @dkuegler can you discuss with @LeHenschel and fix this. If I understand correctly it should be "or .." , but this is not yet fixed in the VINN branch.

PR #181, opened 2022-08-26
Hires-flag for hires processing in recon-surf

LeHenschel:
Added missing hires flag in run-fastsurfer.sh. Will be passed over to recon-surf.sh to allow conforming of images to minimum size (not 1.0 mm).

### DIFF
--- a/FastSurferCNN/run_prediction.py
+++ b/FastSurferCNN/run_prediction.py
@@ -110,6 +110,7 @@ class RunModelOnData:
         self.model = Inference(self.cfg_fin, self.ckpt_fin, args.no_cuda, self.small_gpu)
         self.device = self.model.get_device()
         self.dim = self.model.get_max_size()
+        self.hires = args.hires
 
     def set_view_ops(self, args):
         cfg_cor = set_up_cfgs(args.cfg_cor, args) if args.cfg_cor is not None else None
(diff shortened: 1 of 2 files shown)


m-reuter:
So this passeds hrough the hires flag, but what happens if it is not passed? Will the segmentation be at 1mm to match the recon surf processing? What is the default behavior in Vinn? Maybe we should match the recon surf to that.

LeHenschel:
Yeah, good point. VINN will return the segmentation at the original resolution and recon-surf will fail because the segmentation is not at 1.0 mm. So we should conform the image before passing it to VINN or downsample the hires segmentation.

LeHenschel:
Also these two lines. https://github.com/Deep-MI/FastSurfer/blob/stable/recon_surf/recon-surf.sh#L411 and https://github.com/Deep-MI/FastSurfer/blob/stable/recon_surf/recon-surf.sh#L414 are a bit redundant (if FastSurferVINN/CNN is used before, both the segmentation and orig.mgz exist and are conformed). So the conversion would only do anything if someone is running it with their own segmentation?

LeHenschel:
I updated it so now recon-surf and vinn can run sequentially and will conform the image equally.

Things to discuss:
Currently, default is conforming to 1.0 mm (as in freeSurfer). The hires flag has to be explicitly passed to get hires

m-reuter:
> Also these two lines. https://github.com/Deep-MI/FastSurfer/blob/stable/recon_surf/recon-surf.sh#L411 and https://github.com/Deep-MI/FastSurfer/blob/stable/recon_surf/recon-surf.sh#L414 are a bit redundant (if FastSurferVINN/CNN is used before, both the segmentation and orig.mgz exist and are conformed). So the conversion would only do anything if someone is running it with their own segmentation?

Yes, but theoretically people can run recon-surf on their own images and segmentation. Also those two lines convert inputs to mgz format at the same time.

m-reuter:
Yes, we need to discuss separately if we want to follow FreeSurfer and forcing users to pass the hires flag, or if we want to  determine the best resolution by default (and then have a 1mm or resample flag for those who want to downsample - should not be too many). I think I prefer the latter. Eventually I would like to work in native space (without reslicing to LIA position). But not sure all our FreeSurfer dependencies can handle the shock of not finding FS-conformed inputs :-).

Issue #187, opened 2022-09-09
[optimization] fastsurfervinn interpolating for scale factor 1


dkuegler:
## Feature Description
Create a bypass condition in interpolation_layer.py to skip interpolation, if scale factor is 1

## Motivation
Fastsurfervinn even interpolates if the "corrected" scale factor is 1, i.e. interpolation is effectively an identity operation.

## Additional Context
I already coded the condition and it seems to work, but I have not tested the results. Will do that Monday.

dkuegler:
Adressed in https://github.com/Deep-MI/FastSurfer/pull/188

PR #188, opened 2022-09-12
optimize interpolation_layer.py

dkuegler:
if scale_factor is 1 (1mm resolution), interpolation is the identity operation and my be skipped

Adressing https://github.com/Deep-MI/FastSurfer/issues/187

Speed improvement is about 1% (205 to 203 seconds on cpu)

Tested on bert -- identical for both cpu and gpu

### DIFF
--- a/FastSurferCNN/models/interpolation_layer.py
+++ b/FastSurferCNN/models/interpolation_layer.py
@@ -104,6 +104,16 @@ class _ZoomNd(nn.Module):
         if len(scales_chunks) == 0:
             raise ValueError(f"Invalid scale_factors {scale_factors}, no chunks returned.")
         scales, chunks = map(list, scales_chunks)
+
+        if len(scales) == 1:
+            if isinstance(scales[0], Tensor):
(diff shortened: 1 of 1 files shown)


m-reuter (approved):

m-reuter:
It is checking for all resolutions to be equal to 1 but what if the fixed inner resolution is not 1 mm ? Shouldn't it check if voxel size is equal to inner resolution?

dkuegler:
the inner resolution is unknown to the interpolation layer, it checks, if all samples in the batch have scale_factor 1 (which is the case if `len(scales) == 1` and `scales[0] == ndarray/Tensor(1,1)`. (the scale factor is already the fraction of inner and outer resolution)

So this is a more bare solution that changes the interpolation layer to skip interpolation if none is required.

m-reuter:
Also since speed improvement is negligible I still wonder where the significantly increased run times in VINN come from, if not the interpolation layer.

dkuegler:
I guess we could profile the run times....

m-reuter (approved):

PR #189, opened 2022-09-12
[logging] cleanup and unify the logging in vinn

dkuegler:
cleanup the logging in run_prediction.py
- replace stub get_logger to getLogger
- add `__init__.py` in utils
- rename logging_script.py to logging.py (which is now an extension to the builtin logging, so it is less confusing)
- remove duplicate logging
- move the initialization of the logger to the beginning (instead of in __init__ of RunModelOnData
- added a tqdm progress bar and cleaned log messages for batches in Inference.eval
- change `image.header.get_affine()` to `image.affine` in a couple of places

### DIFF
--- a/FastSurferCNN/data_loader/data_utils.py
+++ b/FastSurferCNN/data_loader/data_utils.py
@@ -23,7 +23,8 @@ from scipy.ndimage import binary_erosion, binary_closing, filters, uniform_filte
 import scipy.ndimage.morphology as morphology
 import nibabel as nib
 import pandas as pd
-import logging
+
+from utils import logging
 from .conform import is_conform, conform, check_affine_in_nifti
 
 
@@ -31,9 +32,7 @@ from .conform import is_conform, conform, check_affine_in_nifti
 # Global Vars
 ##
 SUPPORTED_OUTPUT_FILE_FORMATS = ['mgz', 'nii', 'nii.gz']
-LOGGER = logging.getLogger("eval")
-LOGGER.setLevel(logging.DEBUG)
-LOGGER.addHandler(logging.StreamHandler(stream=sys.stdout))
+LOGGER = logging.getLogger(__name__)
 
 ##
 # Helper Functions
--- a/FastSurferCNN/data_loader/dataset.py
+++ b/FastSurferCNN/data_loader/dataset.py
@@ -24,9 +24,9 @@ import time
 
 import data_loader.data_utils as du
 
-import utils.logging_script as logging
+from utils import logging
 
-logger = logging.get_logger(__name__)
+logger = logging.getLogger(__name__)
 
 
 # Operator to load imaged for inference
--- a/FastSurferCNN/data_loader/loader.py
+++ b/FastSurferCNN/data_loader/loader.py
@@ -21,9 +21,9 @@ import data_loader.dataset as dset
 from data_loader.augmentation import ToTensor, ZeroPad2D, AddGaussianNoise
 import torchio as tio
 
-import utils.logging_script as logging
+from utils import logging
 
-logger = logging.get_logger(__name__)
+logger = logging.getLogger(__name__)
 
 
 def get_dataloader(cfg, mode):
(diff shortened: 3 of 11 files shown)


m-reuter:
Thanks. More than the above have changed, like the new args2config treatment and also how you get the affine. Why has that been changed?

dkuegler:
This is because of
> move the initialization of the logger to the beginning (instead of in init of RunModelOnData

Basically, you want to initialize the logger before you do the first logging. To do that, you have to move the initialization up in order (out of the init method), since there are log messages before that.
If you dont do that, logging does not actually log in the beginning (at least info messages are dropped)

In general, I would consider it best practice, to initialize the "global logger" (or root logger) with `logging.basicConfig` as early as possible and always in the script (i.e. in the `if __name__ == "__main__":` clause) so you always will have exactly one logger.

m-reuter:
I also meant the change from 
`self.orig.header.get_affine()`
to 
`self.orig.affine`
in run_prediction line 242. Or is that essentially the same? Not sure, as one could be a stored matrix, the other could compute that matrix from stored directions and voxel sizes or quaternion representation in a different way.

LeHenschel:
i think this is an issue depending on what kind of file you are loading (nifti or mgz). For one of them, the affine is associated with the header, for the other the affine is associated with the image object. Would be good to have an IO routine for loading as well that checks the type and automatically sets this correctly.

dkuegler:
Yeah, I missed to discus that in the pull request.

get_affine is deprecated over the affine property
https://nipy.org/nibabel/reference/nibabel.spatialimages.html?highlight=get_affine

I changed that because even with nibabel 3.2 I was getting deprecation warnings, but this is not really necessary for this PR, I can revert those changes.

dkuegler:
I think I was loading nifty files, so that might be true, but theoretically, the affine property that I am using now should work indepent of the file type (it is the general interface of nibabel).

Was there a specific reason that you used get_affine of the header?

LeHenschel:
I am not 100 % sure, but I think there was an issue with affine instead of get_affine from header, if you change properties of the image and header, i.e. if you change the scale-factor in the header, this is reflected in the affine from header.get_affine() but not img.affine.

dkuegler:
To clarify, the load function for `nibabel.MGHImage` https://github.com/nipy/nibabel/blob/62aea04248e70d7c4529954ca41685d7f75a0b1e/nibabel/freesurfer/mghformat.py#L572 reads the affine from the header.

Now, since you are able to modify the header in the image object, (which does not change the header's affine) you should NEVER use `header.get_affine()` manually, but always use `image.affine`. This affine also gets written into the header in the save function (happens here https://github.com/nipy/nibabel/blob/62aea04248e70d7c4529954ca41685d7f75a0b1e/nibabel/freesurfer/mghformat.py#L592).

So, we should always use `image.affine` and that also solves the issue of not needing somne sort of "IO routine"

dkuegler:
> I am not 100 % sure, but I think there was an issue with affine instead of get_affine from header, if you change properties of the image and header, i.e. if you change the scale-factor in the header, this is reflected in the affine from header.get_affine() but not img.affine.

Yes, these are not "synchronized" internally. `image.affine` is stored in the image object, `header.get_affine()` reconstructs the affine from the header. As a general rule, if you change one, the other does not change. Also, it is non-trivial to set the affine of the header (you have to call `update_header()`, which calls `_affine2header()`) https://github.com/nipy/nibabel/blob/14db79eae0162fe8a1573acad752d9976d9cab2b/nibabel/spatialimages.py#L476 https://github.com/nipy/nibabel/blob/62aea04248e70d7c4529954ca41685d7f75a0b1e/nibabel/freesurfer/mghformat.py#L622

TL;DR: The intended usage is to always use `image.affine`. If you set the affine manually and want to read something from the header (synchronize header and `image.affine`) call `image.update_header()`.

m-reuter:
That makes sense and I agree. We should move away from deprecated stuff anyway.

m-reuter (approved):

dkuegler:
@m-reuter I beliebe the conform script also still uses `header.get_affine()`...

m-reuter:
Could be. Can you write a short issue to check all occurrences and update those to get a away from deprecated code. Also to test those occurrences for various nifti and mgz inputs (also non - isotropic).

PR #195, opened 2022-09-23
bump to FS7.3.2 and Ubuntu 20

m-reuter:
This PR updates the FreeSurfer download package to FreeSurfer 7.3.2 and uses also the Ubuntu 20 package now.


m-reuter:
I have tested this on one of the local high-res cases both with and without the -hires flag and it finished successfully.

Issue #24, opened 2020-07-24
Error running FastSurfer


nunoloucao:
Hi,

I'm getting this error:
./run_fastsurfer.sh: line 285: syntax error near unexpected token `&'
./run_fastsurfer.sh: line 285: `  date  |& tee -a $seg_log'

When i run the code in the examples.
Can you help me out?

Thanks.
Best,Nuno

LeHenschel:
Hey Nuno,

I guess this is related to issue #7 . What OS are you on and what version of bash are you using?

Best,
Leonie

nunoloucao:
Hey Leonie,

I'll check issue 7.
MacOS 10.15.6 and bash 5.0.18.

Thanks.
Best,
Nuno

m-reuter:
We don't support MacOS yet. Until we do you need to use docker. Sorry.

m-reuter:
Update: in our upcoming release we added support for Mac (Intel cpu-based and M Apple Silicon also GPU on mps device) for the segmentation. Furthermore, we updated install instructions (especially to update the bash shell) and modified recon-surf slightly in #198 to run through. We do not do any testing on the Mac, so you should inspect results.

PR #198, opened 2022-09-28
Feature/qc

m-reuter:
This PR adds:

- a quick_qc.py script for recon-surf.sh to quickly check if the incoming segmentation is severely broken. It checks if the total segmentation volume is less than half of its usual size. A more precise threshold can be set later, once we know better the distribution of correct seg volumes.
- time does not work on a mac, in that case we just skip the timing 
- added a check for the right bash version (> 4.0) which usually is much older on a mac and lead to issues #7 and #24 . Some of those issues could be solved by installing the newer bash and running `bash run_fastsrufer.sh ...` or `bash recon-surf ...`
- fixed binpath if $FASTSURFER_HOME is not set (setting it to ./ did not work, as we are changing directories later in the pipeline). Now it finds the global path of the recon_surf.sh script and uses that directory.

### DIFF
--- a/recon_surf/quick_qc.py
+++ b/recon_surf/quick_qc.py
@@ -1,0 +1,69 @@
+# Copyright 2019 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


m-reuter:
I tested this recon-surf on my Mac and it completed. Timing information is not available but other than that it looks good.

PR #205, opened 2022-11-03
add requirements_cpu.txt and minor changes

agirodi:
This PR adds: 

- an additional requirements_cpu.txt to make a native cpu install for users more accessible.
- changes to the tutorial to use said requirements_cpu.txt
- updates some mentions of dependency from python 3.5 to 3.8
- updates the CONTRIBUTION.md to mention which logs should be provided in an issue


LeHenschel (commented):
Looks overall ok to me. 

One question: Did you test and confirm that the requirements-cpu.txt can be used too successfully install all dependencies, create the docker and run fastsurfer?

LeHenschel:
In the Tutorial:

- Add FastSurferVINN paper as a reference as well
- Same for B2, needs to be updated to FreeSurfer 7. This is already the case for v1.1.1 (https://github.com/Deep-MI/FastSurfer/blob/v1.1.1/recon_surf/recon-surf.sh#L18  --> uses FreeSurfer 7.2 and not FreeSurfer 6.0 as indicated in section B2)
- FAQ section at the end needs to be updated as well (FastSurfer works with FreeSurfer 7.2 now)
- QA-tools part: needs to be checked if this also works with FreeSurfer 7 output (@kdiers)

For VINN (future commit into vinn)
- Version in docker sections is on v1.1.1, this needs to be changed to v2.0.0
- Arguments in run_fastsurfer.sh changed and need to be updated (no_cuda e.g.)

m-reuter:
Currently we test freesurfer 7.3.2 not 7.2 but this is not yet finally decided

LeHenschel:
Yes, but this is the update for v1.1.1/dev, right? Not vinn. We use 7.2.0 there (https://github.com/Deep-MI/FastSurfer/blob/dev/recon_surf/recon-surf.sh#L18)

PR #207, opened 2022-11-09
optimize normalizeWM and mask creation

dkuegler:
- replace repeated percentiles (sorting) with single call
- remove some of the repeated conversion between datatypes (unit/float32)

### DIFF
--- a/recon_surf/N4_bias_correct.py
+++ b/recon_surf/N4_bias_correct.py
@@ -18,12 +18,12 @@
 
 # IMPORTS
 import optparse
-import numpy as np
 import sys
-import SimpleITK as sitk
-import nibabel as nib
-import image_io as iio
 
+import SimpleITK as sitk
+import numpy as np
+
+import image_io as iio
 
 HELPTEXT = """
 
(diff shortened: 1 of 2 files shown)


dkuegler:
There is a bunch of comments and still one debug line in there that need to be cleaned

dkuegler:
My experiments show about 2% speed improvement (for the normalizeWM function it is actually about 50%), but the lionshare of the execution is in the BiasFieldCorrection itself (i.e. in SimpleITK), there is not really anything that CAN be optimized easily short of a. changing the parameters (iterations, levels, etc.) and b. optimizing N4 itself and porting it to pytorch (GPU).

Results tested for one subject. 11 voxels different, max difference 1 (intensity difference of 1)

dkuegler:
reduce_aseg_to_mask Tested on one subject
20%-30% faster
identical results

m-reuter:
thanks for these improvements. Code looks good. @AhmedFaisal95 can you test this on 2-3 cases if results are identical? then we can merge it. Thanks

af-a:
> thanks for these improvements. Code looks good. @AhmedFaisal95 can you test this on 2-3 cases if results are identical? then we can merge it. Thanks

Tested on three cases (1.0mm, 0.8mm, and 0.7mm), the outputs of `reduce_to_aseg.py` and `N4_bias_correct.py` are **identical** according to DICE scores (1.0 in all labels; up to three decimal places).

The `reduce_to_aseg.py` modification decreases run-time by approximately 50%, 36% and 37% in the respective cases.
The effect of the `N4_bias_correct.py` modification on run-time is less significant: 2.3%, 0.4%, -1.4%.

PR #208, opened 2022-11-15
Interface and FastSurfer basepath changes

dkuegler:
Interface changes in preparation for modules.
Cleanup of the import statements.
Change of the "base path" of FastSurferCNN (now the include statements start with FastSurferCNN), i.e. the python search path should include the FastSurfer project folder and not the FastSurferCNN subfolder.

This specifically prepares the change of the run_fastsurfer.sh main entrypoint.

### DIFF
--- a/FastSurferCNN/__init__.py
+++ b/FastSurferCNN/__init__.py
@@ -1,2 +1,0 @@
-import os, sys
-sys.path.append(os.path.abspath(os.path.pardir))
--- a/FastSurferCNN/data_loader/augmentation.py
+++ b/FastSurferCNN/data_loader/augmentation.py
@@ -17,8 +17,6 @@
 # IMPORTS
 import numpy as np
 import torch
-import torchio as tio
-from scipy.ndimage import distance_transform_edt as eucl_distance
 
 
 ##
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -16,6 +16,7 @@
 # IMPORTS
 import argparse
 import sys
+
 import numpy as np
 import nibabel as nib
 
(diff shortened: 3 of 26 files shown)


dkuegler:
I have not tested all scripts yet, so that should be done "soon"-ish.

dkuegler:
I cleaned up the commit history and rebased onto the current Deep-MI:feature/vinn

dkuegler:
I wish I could clean up the history here. github repeatedly did not want to accept my rebase so we have lots of commit messages now. I reset everything and merged instead, at least in the commit history it will be almost clean.

-- Well that kind of happened automatically.

dkuegler:
Rebase and push of the first draft.

LeHenschel:
We should add the PYTHONPATH in the Dockerfile because currently the build fails (download_checkpoints.py does not find FastSurferCNN module).

EDIT: The Dockerfile for FastSurferCNN (segonly) needs to be changed. With the new module structures, the checkpoint download fails (independent of the PYTHONPATH). Reason is, that we only copy FastSurferCNN folder, not the entire thing. I changed the Dockerfile_FastSurferCNN in my aseg_brainmask_module branch (based on this one). @AhmedFaisal95 , @agirodi please check and make sure that the other Docker builds run successfully with the new import statements.

LeHenschel:
Few more errors:
- with the torch version in requirements, mps does not exist as a callable module yet (AttributeError: module 'torch.backends' has no attribute 'mps')
- The checkpoints are downloaded by root in the docker, hence running docker as a different user does not allow access to them (this line from the Dockerfile: RUN cd /fastsurfer ; python3 FastSurferCNN/download_checkpoints.py --all)
- remove_suffix function does not work correctly. If a path to the whole image is passed, nothing happens, because this should remove the filename as well. So either, one has to pass the entire image name as well, or we need to change it, to also remove the last filename. 
- in run_fastsurfer.sh --remove_suffix is not passed, hence the subject name is not set correctly

I fixed the last two points in the #212 pull request

dkuegler:
> - with the torch version in requirements, mps does not exist as a callable module yet (AttributeError: module 'torch.backends' has no attribute 'mps')

Best solution just use a "newer" Pytorch. Why would we want to use an old version anyways?

> - The checkpoints are downloaded by root in the docker, hence running docker as a different user does not allow access to them (this line from the Dockerfile: RUN cd /fastsurfer ; python3 FastSurferCNN/download_checkpoints.py --all)

we can just make them world readable `RUN cd /fastsurfer ; python3 FastSurferCNN/download_checkpoints.py --all; chmod w+r -R checkpoints` I think. @AhmedFaisal95

agirodi:
> EDIT: The Dockerfile for FastSurferCNN (segonly) needs to be changed. With the new module structures, the checkpoint download fails (independent of the PYTHONPATH). Reason is, that we only copy FastSurferCNN folder, not the entire thing

The other Dockerfiles at least need the definition of the Pythonpath and the Dockerfile_FastSurferCNN_CPU needs the same changes to the copys as the GPU file. The at least build, but I haven't gotten to run them due to the mentioned checkpoint download by root.

PR #212, opened 2022-12-05
Feature/aseg brainmask module

LeHenschel:
This pull requests moves brainmask and aseg generation to the segmentation pipeline. Branch is based on Davids feature/interface pull request and should be merged afterwards.

- reduce_aparc_to_aseg calls were moved from recon-surf.sh to run_prediction.py
- the file is for now still in recon_surf (./recon_surf/reduce_to_aseg.py), as we want to reorder the directories anyhow in the future. Would make more sense to have this in CNN.
- sid, brainmask_name and aseg_name added as available argumentparser arguments
- separate sid flag and update of --remove_suffix documentation to assure subject_name is set correctly
- run_fastsurfer.sh passes sid to run_prediction.py
- Dockerfile for running segmentation only on GPU was adopted by exporting PYTHONPATH and including all relevant python files.

Changes were tested on bert and IXI626. @agirodi, @AhmedFaisal95 please pull the branch:
1. run tests on the validation set (full pipeline)
2. update all dockerfiles and check that they all (!) work correctly dockerfiles
3. check that the downloaded checkpoints inside the docker file are accessible to all users, not just root (as is currently the case)


FastSurferCNN/data_loader/conform.py
```diff
@@ -238,7 +239,7 @@ def conform(img, order=1, conform_min=False):
     format and enforces 1 mm isotropic voxel sizes.
     Difference to mri_convert -c is that we first interpolate (float image), and then rescale to uchar. mri_convert is
     doing it the other way. However, we compute the scale factor from the input to be more similar again
-    :param nibabel.MGHImage img: loaded source image
```
dkuegler: MGHImage does not actually include NiftyImage, it would be better to only assume the more general SpatialImage (I think this also has a header and an affine, maybe we need to figure out which feature we need that is not included in SpatialImage).

FastSurferCNN/run_prediction.py
```diff
@@ -31,6 +31,7 @@
 from FastSurferCNN.utils.misc import find_device
 from FastSurferCNN.data_loader import data_utils as du, conform as conf
 from FastSurferCNN.quick_qc import check_volume
+import recon_surf.reduce_to_aseg as rta
```
dkuegler: can we move `reduce_to_aseg.py` into FastSurferCNN then ? I think right now there are already the `conform.py` dependencies in recon-surf and cyclic dependencies should be avoided.

dkuegler (commented):
The MGHImage comment is a bit unclear, maybe that is even #208 stuff.

FastSurferCNN/data_loader/conform.py
LeHenschel: I changed it. We should also consider making the function more general (i.e. at the moment only MGHImage can/will be returned).

FastSurferCNN/run_prediction.py
LeHenschel: Yeah, sure. As stated in the description, I left it in recon-surf for now as we anyhow want to restructure the directories. But I can reorder it. Will make the Dockerfile easier as well

LeHenschel: reduce_to_aseg.py does not import any modules from CNN though

dkuegler: Neither does it from recon_surf.
This really is not a critical issue, but in general cyclic dependencies should be avoided by design.

Issue #190, opened 2022-09-14
replace deprecated image.header.get_affine with image.affine


dkuegler:
## Feature Description
Search and replace `get_affine()` by `image.affine`

Both master and feature/vinn use it at least in `FastSurferCNN/data_loader/conform.py`.
In general, best practice is to always use `image.affine`, since header and `image.affine` only get synchronized during initialization and when `image.update_header()` is called (this updates the header by pushing the `image.affine` into the header).

Probably also move the conversion from any format to mgh into a dedicated function.

## Motivation
`image.header.get_affine()` and `image.get_affine()` is deprecated by nibabel since version 2.0

See also https://github.com/Deep-MI/FastSurfer/pull/189

m-reuter:
After updating it should be tested with both mgz and nii inputs and with various headers (e.g. non-isotropic voxels)

dkuegler:
Currently, `get_affine` is only used in `conform.py`. However, `conform.py` really only supports the output of `MGHImage`s, which is a special case for the affine calculation.
In essence, our code right now internally requires `MGHImage`s and `MGHHeader`s to work properly.
Ideally, we would adapt instances, where we "explicitly" create `MGHImage`s to be more versatile.

PR #213, opened 2022-12-07
Small doc changes and organization in conform.py

dkuegler:
Update docstrings
pull out the get_affine() into the code and add documentation w.r.t the deprecated get_affine function. 
Add Typing

See #190

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -14,6 +14,8 @@
 
 
 # IMPORTS
+import logging
+from typing import Optional, Type, Tuple, Union
 import argparse
 import sys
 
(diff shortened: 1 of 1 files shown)


FastSurferCNN/data_loader/conform.py
```diff
@@ -232,17 +257,24 @@ def findMinSizeConformDim(img, max_size=1, min_dim=256):
     return min_size, conform_dim
 
 
-def conform(img, order=1, conform_min=False):
+def conform(img: nib.analyze.SpatialImage, order: int = 1, conform_min: bool = False, dtype: Optional[Type] = None) -> nib.MGHImage:
```
m-reuter, line 266: why has dtype been added, when it is neither used in the code below (nor documented in the help text)?

dkuegler: Ups. Incomplete. I wanted to add the ability to store float images so we get rid of rounding errors.... But that was not completed.

Issue #109, opened 2022-02-25
Support FastSurferCNN on Apple M1


emilljungberg:
I'm trying to run FastSurfer CNN for segmentation on my MacBook Pro with the Apple M1 chip. I was able to build to docker container for CPU using your instructions (example 4) by adding in `--platform linux/x86_64`, but there seems to be some problem running the program. When I execute the following (note the added platform here as well)
```
docker run --platform linux/x86_64 -v $SUBJECT_DIR:/data \
           -v $outpath:/output \
           --rm --user 501 fastsurfercnn:cpu \
           --i_dir /data \
           --in_name mri/mprage.mgz \
           --o_dir /output \
           --out_name mri/aparc.DKTatlas+aseg.deep.mgz \
           --log deep_surfer.log \
           --no_cuda
```
I get the following output
```
Total Dataset Size is 1
Running Fast Surfer on sub01
Reading volume /data/sub01/mri/mprage.mgz
Conforming image to UCHAR, RAS orientation, and 1mm isotropic voxels
Input:    min: 0  max: 1851
rescale:  min: 0.0  max: 831.099  scale: 0.30682265289694727
Output:   min: 0.0  max: 255.0
Cuda available: False, # Available GPUS: 0, Cuda user disabled (--no_cuda flag): True, --> Using device: cpu
```
The program creates the correct output folder structure but there is no output files in there. Any clue what might be going wrong?

## Environment
 - FastSurfer Version: Pulled from Github 2022/02/25
 - FreeSurfer Version: 7.2.0 (know this is higher than recommended but not used for the segmentation right?)
 - OS: MacOS 12.1
 - CPU: Apple M1 Pro

m-reuter:
Hi Emil, 
we do not support the M chips yet. Probably in the future. Maybe it is easy or maybe not. We can keep this issue open, for people to post if they have success getting it running. 

And also 7.2 is not only not recommended, it is not supported. Use 6.0 for now. But you are right, for only the segmentation it is not used and should not matter.

LeHenschel:
Maybe also a memory issue (see issue https://github.com/Deep-MI/FastSurfer/issues/40)? In order to run FastSurfer you need around 10 GB RAM, but Docker Desktop on MAC is set to use 2 GB runtime memory by default. You can overwrite the settings under Docker Desktop --> Preferences --> Resources --> Advanced, slide the bar under Memory to 10 GB; see: https://docs.docker.com/docker-for-mac/ for details.

[...]

PR #215, opened 2022-12-09
Create a docker multistage build script

dkuegler:
Create a docker multistage build script that is compatible with FastSurfer interface rework.

All images are built with the "core" Dockerfile at Docker/Dockerfile. Also update some additional folders to not be copied automatically by the build (.dockerignore).

TODOs:
- Testing
- check for the updates to AMD (rocm/pytorch is now python 3.8 as well)
-


dkuegler:
@AhmedFaisal95 Please have a look and see what still needs to be done.
I have already added some documentation in Docker/README.md

At a minimum, this drastically reduces the doubling of code. 

It seems that DOCKER_BUILDKIT also supports the script now. Which speeds up builds a lot, since it builds stages that are independent in parallel.

For example
```
DOCKER_BUILDKIT=1 docker build --rm=true --target runtime --build-arg DEVICE=cuda --build-arg FREESURFER=pruned -t fastsurfer:gpu -f ./Docker/Dockerfile .
```

af-a:
This seems to run fine except when:
1. build arg `DEVICE` is set to `cpu` (any target) 
2. the target is `runtime_surf_only` and `DEVICE` is set to `none`.

### Case 1
Command:
```
DOCKER_BUILDKIT=1 docker build --rm=true --target runtime --build-arg DEVICE=cpu -t IMAGE_ID -f ./Docker/Dockerfile .
```

Result:
```
[+] Building 0.6s (11/17)                                                                                                                                                                                                                                                                                                                                                           
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                                                                                                           0.3s
 => => transferring dockerfile: 5.62kB                                                                                                                                                                                                                                                                                                                                         0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                                                                                                              0.4s
 => => transferring context: 136B                                                                                                                                                                                                                                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                                                                                                                                                                                                                                                                                                                0.0s
 => CANCELED [internal] load build context                                                                                                                                                                                                                                                                                                                                     0.0s
 => [build_conda_base 1/3] FROM docker.io/library/ubuntu:20.04                                                                                                                                                                                                                                                                                                                 0.0s
 => CACHED [build_conda_base 2/3] RUN apt-get update && apt-get install -y --no-install-recommends       wget       git       ca-certificates       upx       file &&     apt clean &&     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*                                                                                                                                       0.0s
 => CACHED [build_conda_base 3/3] RUN wget --no-check-certificate -qO ~/miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-py38_4.11.0-Linux-x86_64.sh  &&      chmod +x ~/miniconda.sh &&      ~/miniconda.sh -b -p /opt/conda &&      rm ~/miniconda.sh                                                                                                             0.0s
 => CACHED [build_conda_cpu 1/1] RUN conda install --channel pytorch cpuonly=1.0                                                                                                                                                                                                                                                                                               0.0s
 => CACHED [runtime 1/7] RUN apt-get update && apt-get install -y --no-install-recommends       tcsh       time       bc       gawk       libgomp1       libquadmath0 &&     apt clean &&     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*                                                                                                                                    0.0s
 => CACHED [runtime 2/7] RUN echo "source /venv/bin/activate" >> ~/.bashrc                                                                                                                                                                                                                                                                                                     0.0s
 => ERROR [runtime 3/7] COPY --from=build_conda /venv /venv                                                                                                                                                                                                                                                                                                                    0.0s
------
 > [runtime 3/7] COPY --from=build_conda /venv /venv:
------
failed to compute cache key: "/venv" not found: not found
```


### Case 2
Command:
```
DOCKER_BUILDKIT=1 docker build --rm=true --target runtime_surf_only --build-arg DEVICE=none --build-arg FREESURFER=pruned -t IMAGE_ID -f ./Docker/Dockerfile .
```

Result:
```
[+] Building 1.6s (4/4) FINISHED                                                           
 => [internal] load build definition from Dockerfile                                  0.3s
 => => transferring dockerfile: 140B                                                  0.0s
 => [internal] load .dockerignore                                                     0.5s
 => => transferring context: 136B                                                     0.0s
 => ERROR [internal] load metadata for docker.io/library/base_none:latest             1.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                       0.0s
------
 > [internal] load metadata for docker.io/library/base_none:latest:
------

```

af-a:
The issues mentioned above should be fixed in this PR: https://github.com/dkuegler/FastSurfer/pull/2

dkuegler:
@AhmedFaisal95 Can you confirm this works now?

We need a "basic test" of all images :).

dkuegler:
@m-reuter The Apple M1 device is missing from the docker multistage built. The reason is that it was not in `dev` at all.

For the AMD build, the base image got updated and I am not really sure if the updated image works. Maybe, someone can check this as well?

af-a:
@dkuegler 

> @AhmedFaisal95 Can you confirm this works now?
> 
> We need a "basic test" of all images :).

I already tested the fixes before submitting the PR, specifically all images except for the AMD case (no hardware).

dkuegler:
The build script should also compile the byte code https://docs.python.org/3/library/py_compile.html .

dkuegler:
I think what is left for this PR to be merged is proper testing.
Especially for the AMD build.

m-reuter:
> The Apple M1 device is missing from the docker multistage built. The reason is that it was not in `dev` at all.

For M1 we do not use a docker, only native install on Mac. Inside Docker it was much slower for yet unknown reasons, see #109

dkuegler:
@agirodi can you test the AMD docker script?

agirodi:
> @agirodi can you test the AMD docker script?

I just did and it is working correctly. The build went without issues and running produced a correct segmentation.

dkuegler:
Then, this PR should be ready for merging.

m-reuter:
A few points for documentation:
- due to the minimal larger size when including FreeSurfer binaries, we will stop publishing docker images without FreeSurfer
- pytorch gpu is still much larger than cpu version so we will continue to publish both versions
- AMD is experimental and can break without notice, which is why we will provide build files, but will not publish the container for now. 
- As dependencies (versions) remain relatively stable, we will have a base release image without pytorch (including all other dependencies like python packages and FreeSurfer binaries), and then two versions from that with pytorch GPU and CPU. 
- FastSurfer is then added on top of these three as the final layer. The non-pyorch entry point remains recon-surf, for the other two it will be run_fastsurfer.sh.. Users who want to directly run_prediction, need to modify the docker run command and overwrite the entry point or add another layer. 

The advantage of this approach is, that for future docker builds we can pull the base release images and use them (in most cases they probably exist locally anyway). Adding FastSurfer is then extremely fast :-)

Docker/Dockerfile
```diff
@@ -1,102 +1,201 @@
...
+
+
+# =========================================
+# Here, we create the smaller runtime image
+# =========================================
+FROM $RUNTIME_BASE_IMAGE AS runtime
 
-# Install required packages for freesurfer to run
+ENV LANG=C.UTF-8 \
+    HSA_OVERRIDE_GFX_VERSION=10.3.0
```
m-reuter: maybe we need a comment here, why this is needed? And when it would need to be changed to a new version?

Docker/README.md
```diff
@@ -42,35 +43,42 @@ All other available flags are identical to the ones explained on the main page [
...
+python build.py --device cuda
 ```
 
 For running the analysis, the command is basically the same as above for the prebuild option:
 ```bash
 docker run --gpus all -v /home/user/my_mri_data:/data \
                       -v /home/user/my_fastsurfer_analysis:/output \
                       -v /home/user/my_fs_license_dir:/fs_license \
-                      --rm --user $(id -u):$(id -g) my_fastsurfer:gpu \
+                      --rm --user $(id -u):$(id -g) my_fastsurfer:2.2.1 \
```
m-reuter: We did not want version numbers hard coded in the documentation.

Docker/README.md
```diff
@@ -83,16 +91,15 @@ docker run --gpus all -v /home/user/my_mri_data:/data \
...
+python build.py --device cpu
 ```
 
 For running the analysis, the command is basically the same as above for the GPU option:
 ```bash
 docker run -v /home/user/my_mri_data:/data \
            -v /home/user/my_fastsurfer_analysis:/output \
            -v /home/user/my_fs_license_dir:/fs_license \
-           --rm --user $(id -u):$(id -g) my_fastsurfer:cpu \
+           --rm --user $(id -u):$(id -g) my_fastsurfer:2.2.1-cpu \
```
m-reuter: another hard coded version

Docker/README.md
```diff
@@ -191,7 +128,7 @@ docker run --rm --cap-add=SYS_PTRACE --security-opt seccomp=unconfined \
 	   --shm-size 8G \
 	   -v /home/user/my_mri_data:/data \
 	   -v /home/user/my_fastsurfer_analysis:/output \
-	   my_fastsurfer:gpu-amd \
+	   my_fastsurfer:2.2.1-rocm \
```
m-reuter: another hard coded version

Docker/Dockerfile
dkuegler: This is a flag for AMD ROCm. It should probably not be part of the "standard" docker.
In fact, since this is a flag that enables ROCm support on hardware that AMD does not officially support ROCm for, it should probably not be part of the docker container at all.
I suggest this flag **has to be passed** by the user `docker run ... -e HSA_OVERRIDE_GFX_VERSION=10.3.0 fastsurfer:rocm ...` instead of blanket-setting it here. Because other hardware might need this variable to be "unset".

m-reuter: yes, makes sense. It should be added to the documentation (README). I think our AMD GPU needed this flag and probably other consumer cards.

PR #216, opened 2022-12-09
discourage running fastsurfer as root

dkuegler:
Running Fastsurfer as root can cause a couple of problems with files and folders being created as root. 

This PR discourages running Fastsurfer as root by 
1. Exiting with an error Message if it is done.
2. Introducing an --allow_root flag in run_fastsurfer.sh to overwrite this behavior.

### DIFF
--- a/run_fastsurfer.sh
+++ b/run_fastsurfer.sh
@@ -49,6 +49,7 @@ doParallel=""
 run_aparc_module="1"
 threads="1"
 python="python3.8"
+allow_root="0"
 
 # Dev flags defaults
 vcheck=""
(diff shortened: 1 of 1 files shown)


dkuegler:
Question. Should similar checks be done in recon-surf/recon-surf.sh and FastSurferCNN/run_prediction.py?

Also. I think this should be fine, but the code is not actually tested.

m-reuter:
I think it would be good to add it to those as well, at least to all entry points of any docker image. Also in the error message it would be good to only print one option ( the -u flag) followed by 'id -u' to avoid users specifying their use name.

dkuegler:
I added the requested changes.

dkuegler:
@m-reuter unfortunately I cannot reopen this PR and have to create a new PR it seems.

PR #223, opened 2022-12-14
Feature/set t1 to default if empty

LeHenschel:
Change behaviour of run_fastsurfer.sh
- t1 does not need to be defined when running the surface pipeline only. In this case a conformed image should/can already exist from running the segmentation pipeline. Check if this exists, is already done
- call for seg-only simplifies to defining --sd and --sid when segmentation was already generated with default arguments for --conformed_name and --aparc_aseg_segfile

Fix error in recon-surf.sh
- hires flag was not turned on, when recon-surf was called directly without any argument to --vox_size. Hires flag is now turned on when --vox_size is set to auto, independent if this was explicitly passed as a flag or not (vox_size is set to auto by default).

### DIFF
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -288,7 +288,7 @@ case $key in
     ;;
     --hires)
     echo "WARNING: --hires is deprecated, use --vox_size auto."
-    vox_size=1
+    vox_size="auto"
     hiresflag="-hires"
     shift # past argument
     ;;
@@ -545,7 +545,7 @@ cmin=""
 if  [ "$vox_size" == "auto" ]
 then
   cmin="--conform_min"
-  hiresflag="--hires"
+  hiresflag="-hires"
 fi
 cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $t1 --check_only $cmin --verbose"
 RunIt "$cmd" $LF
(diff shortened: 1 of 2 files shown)


m-reuter:
Not sure about the hires stuff. What if Auto results in a 1mm image? We do not want hires then. Also David is working on a fix, and this will probably conflict.

LeHenschel:
The current behaviour is that auto = hires. The check that david is implementing will be added to run_fastsurfer.sh not recon-surf. We talked about removing the flag in recon-surf and add only the check for conform min/hires with the conform.py --check_only. 

As currently, auto=hires and this is set on by default in recon-surf, but the -hires flag is not turned on, this effectively negates this (run with 1.0 mm settings in recon-surf for all recon-all runs, but hires for all other runs). So I would suggest to merge this request now to have consistent behaviour until David finishes his changes. And then it will be a simple merge/rebase.

dkuegler:
@LeHenschel see #225 for my changes and this impacts both recon-surf and run_fastsurfer...

LeHenschel:
Yes,  I am aware of that @dkuegler. See my comment above

> The check that david is implementing will be added to run_fastsurfer.sh not recon-surf. We talked about removing the flag in recon-surf and add only the check for conform min/hires with the conform.py --check_only.

= changes in run_fastsurfer.sh for auto flag and in recon-surf for conform

PR #225, opened 2022-12-15
 Voxel size passing and hires pipeline selection changes to allow consistent processing

dkuegler:
Interface change to conform.py
- Adds new flag: --conform-size, analogeous to mri_convert --conform_size
Interface change to recon-surf.sh, run_prediction.py and run_fastsurfer.sh.
- Remove --vox_size from recon-surf.sh, vox-size is automatically extracted from the image file
- Add additional options to --vox_size in run_prediction.py and run_fastsurfer.sh:
  Now accepts a voxel size between 0.7 and 1 (which forces that specific voxel size)
  or 'min' (default), which performs conforming to the minimal voxel size (akin to mri_convert --conform_min).
  Note: the naming 'auto' has been removed in favor of 'min'.
  Lower than 0.7 prints a experimental warning.
  Higher than 1 exists with an error.

The function in conform.py have been adapted/extended to allow the passing of the 'target voxel size'.
The conform_min argument was therefore superceeded by a conform_vox_size argument, which can also be 'min'
to get the conform_min behavior.

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -15,17 +15,22 @@
 
 # IMPORTS
 import logging
-from typing import Optional, Type, Tuple, Union
+from typing import Optional, Type, Tuple, Union, overload, Dict, Literal
 import argparse
 import sys
 
 import numpy as np
 import nibabel as nib
 
+from FastSurferCNN.utils.arg_types import (vox_size as __vox_size, target_dtype as __target_dtype,
+                                           conform_to_one as __conform_to_one, VoxSizeOption)
+
 HELPTEXT = """
 Script to conform an MRI brain image to UCHAR, RAS orientation, and 1mm or minimal isotropic voxels
 USAGE:
 conform.py  -i <input> -o <output> <options>
+OR
+conform.py  -i <input> --check_only <options>
 Dependencies:
     Python 3.8
     Numpy
@@ -54,13 +59,24 @@ def options_parse():
     parser.add_argument('--check_only', dest='check_only', default=False, action='store_true',
                         help='If True, only checks if the input image is conformed, and does not return an output.')
     parser.add_argument('--seg_input', dest='seg_input', default=False, action='store_true',
-                        help='Specifies whether the input is a seg image. If true, '
-                             'the check for conformance disregards the uint8 dtype criteria')
+                        help='Specifies whether the input is a seg image. If true, the check for conformance '
+                             'disregards the uint8 dtype criteria. Use --dtype any for equivalent results. '
+                             '--seg_input overwrites --dtype arguments.')
+    parser.add_argument('--vox_size', dest='vox_size', default=1., type=__vox_size,
+                        help="Specifies the target voxel size to conform to. Also allows 'min' for conforming to the "
+                             "minimum voxel size, otherwise similar to mri_convert's --conform_size <size> "
+                             "(default: 1, conform to 1mm).")
     parser.add_argument('--conform_min', dest='conform_min', default=False, action='store_true',
                         help='Specifies whether the input is or should be conformed to the '
-                             'minimal voxel size (used for high-res processing)')
-    parser.add_argument('--dtype', dest="dtype", default="uint8",
-                        help='Specifies the target data type of the target image (default: uint8, as in FreeSurfer)')
+                             'minimal voxel size (used for high-res processing) - overwrites --vox_size.')
+    advanced = parser.add_argument_group("Advanced options")
+    advanced.add_argument('--conform_to_1_threshold', type=__conform_to_one,
+                          help="Advanced option to change the threshold beyond which images are conformed to 1"
+                               "(default: infinity, all images are conformed to their minimum voxel size).",
+                          )
+    parser.add_argument('--dtype', dest="dtype", default="uint8", type=__target_dtype,
+                        help="Specifies the target data type of the target image or 'any' (default: 'uint8', "
+                             "as in FreeSurfer)")
     parser.add_argument('--verbose', dest='verbose', default=False, action='store_true',
                         help='If verbose, more specific messages are printed')
     args = parser.parse_args()
@@ -70,6 +86,8 @@ def options_parse():
         sys.exit('ERROR: Please specify output image')
     if args.check_only and args.output is not None:
         sys.exit('ERROR: You passed in check_only. Please do not also specify output image')
+    if args.seg_input and args.dtype not in ["uint8", "any"]:
+        print("WARNING: --seg_input overwrites the --dtype arguments.")
     return args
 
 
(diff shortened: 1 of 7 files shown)


dkuegler:
I still have to test the changes.

FastSurferCNN/utils/parser_defaults.py
```diff
@@ -90,10 +102,10 @@ def _stub(parser):
         help="Absolute path to file in which a list of subjects that failed QC check  (when processing multiple "
              "subjects) will be saved. If not set, the file will not be saved."),
     "vox_size": __arg(
-        '--vox_size', choices=["auto", "1"], default="auto", dest='vox_size',
-        help="Choose the primary voxelsize to process, must be either '1' or 'auto'. '1' forces "
-             "processing at 1mm voxel size, 'auto' enables processing at higher resolutions (conforming"
-             " to the smallest voxel edge length). Select 'auto' (default) for hires processing."),
+        '--vox_size', type=__vox_size, default="min", dest='vox_size',
```
LeHenschel, line 93: Should this not include a disclaimer that if the res is above 0.95 it will be set to 1.0 mm with min?

recon_surf/recon-surf.sh
```diff
@@ -540,18 +524,55 @@ echo " " |& tee -a $LF
 echo "================== Creating orig and rawavg from input =========================" |& tee -a $LF
 echo " " |& tee -a $LF
 
+CONFORM_LF=$SUBJECTS_DIR/$subject/scripts/conform.log
```
LeHenschel, line 521: what is the rationale about having this in a separate file? I would find it beneficial to have this information in recon-surf.log right at the start.

LeHenschel:
The changes to recon-surf.sh seem overly complex to me. 

a.) hires_voxsize_threshold=0.999 is fixed in the beginning. This can not be changed via the interface. Why not have a vox-size threshold default argument in the function call to conform.py --check-only? This way this does not need to be present in recon-surf.

b.) Why have the exact vox-size written to a separate file? Why not just a boolean value (or sth. like "Conform-min", "1"). And then check for just that one value + set conform-min/hires flag accordingly?  If this is about making sure the resolutions between segmentation and orig are equal:
    1.) Change the call for conform.py --check-only (=option to pass in multiple files, check all of them for conformance, return error if a. they are not conform, b. they have differing resolutions)
    2.) latest when mask and image are passed to a call the program will crash
    3.) Could be a simple python script or inline call that compares the header voxelsize (header_img.get_voxsize() == header_orig.get_voxsize()

FastSurferCNN/data_loader/conform.py
```diff
@@ -277,25 +302,25 @@ def conform(img: nib.analyze.SpatialImage, order: int = 1, conform_min: bool = F
     Args:
         img: loaded source image
         order: interpolation order (0=nearest,1=linear(default),2=quadratic,3=cubic)
-        conform_min: conform image to minimal voxel size (for high-res)
+        conform_vox_size: conform image the image to voxel size 1. (default), a
```
m-reuter, line 309: conform image the image ... ( duplicate "image")

recon_surf/recon-surf.sh
```diff
@@ -286,24 +283,10 @@ case $key in
     fssurfreg=1
     shift # past argument
     ;;
-    --hires)
-    echo "WARNING: --hires is deprecated, use --vox_size auto."
-    vox_size=1
-    hiresflag="-hires"
+    --hires|--vox_size)
```
m-reuter: since we are changing the interface for 2.0 release anyway, I think this can be removed and an error given if someone passes any unknown parameter option.

FastSurferCNN/utils/parser_defaults.py
dkuegler: Well, all that conforming did not actually happen yet :(
So I think I added that now. I'll have to test the code now.

dkuegler:
Unfortunately, there are some nasty dependencies in the code that make many easy solutions non-viable.

There are 4 interfaces that all need to have a sensible behavior.
1. conform.py
2. run_fastsurfer.sh
3. recon-surf.sh
4. run_prediction.py

dkuegler:
I cleaned up the recon-surf file as well. I usually code on the side of clear call interfaces and good error handling, that kind of goes out the window here, but should probably still work with the changed (shorter) structure -- and different interface in conform.py

FastSurferCNN/data_loader/data_utils.py
```diff
@@ -53,8 +53,9 @@ def load_and_conform_image(img_filename, interpol=1, logger=LOGGER, conform_min
     :return: nibabel.MGHImage orig: conformed image
     """
     orig = nib.load(img_filename)
-
-    if not is_conform(orig, conform_min=conform_min):
+    # is_conform and conform accept numeric values and the string 'min' instead of the bool value
+    _conform_vox_size = 'min' if conform_min else 1.
+    if not is_conform(orig, conform_vox_size=_conform_vox_size):
```
LeHenschel, line 58: Why is it not possible to conform to a given size in this function? Everywhere else this is possible, only here it is min or 1

recon_surf/recon-surf.sh
```diff
@@ -540,18 +518,31 @@ echo " " |& tee -a $LF
...
+# look into the CONFORM_LF to find the voxel sizes, the second conform.py call will check the legality of vox_size
+vox_size=`cat $CONFORM_LF | grep -E " - Voxel Size " | cut -d' ' -f5 | cut -d'x' -f1`
+# remove the temporary conform_log (all info is also in the recon-surf logfile)
+if [ -f "$CONFORM_LF" ]; then rm -f $CONFORM_LF ; fi
+
+# here, we check the correct vox_size by passing it to the next conform, so errors in this line might be caused above
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $aparc_aseg_segfile --check_only --vox_size $vox_size --dtype any --verbose"
+RunIt "$cmd" $LF
+
+if [ "$vox_size" -lt "$hires_voxsize_threshold" ]
```
LeHenschel, line 538: Why is this if block necessary? Does not really make sense to me

recon_surf/recon-surf.sh
```diff
@@ -641,7 +632,7 @@ if [ "$get_t1" == "1" ]
 then
   # create T1.mgz from nu (!! here we could also try passing aseg?)
   cmd="mri_normalize -g 1 -seed 1234 -mprage $mdir/nu.mgz $mdir/T1.mgz"
-  if [ "$vox_size" == "auto" ]
+  if [ "$vox_size" -lt "$hires_voxsize_threshold" ]
```
LeHenschel: Didn't we want to get rid of this and just return True/False for voxsize instead of the actual voxel size? And have the check with the hires_voxsize_threshold in conform.py

recon_surf/recon-surf.sh
```diff
@@ -540,18 +518,31 @@ echo " " |& tee -a $LF
...
 echo " " |& tee -a $LF
 
+CONFORM_LF=$SUBJECTS_DIR/$subject/scripts/conform.log
+if [ $CONFORM_LF != /dev/null ] ; then  rm -f $CONFORM_LF ; fi
+echo "Log file for Conform test" > $CONFORM_LF
+
 # check for input conformance
-cmin=""
-if  [ "$vox_size" == "auto" ]
+cmd="$python ${binpath}../FastSurferCNN/data_loader/conform.py -i $t1 --check_only --vox_size min --verbose"
```
LeHenschel, line 526: Why is min passed here, but in the call below we pass the exact voxelsize? To check for conformity between the two?

recon_surf/recon-surf.sh
LeHenschel: or at least also set the flags for conform min and hires here, so we can avoid the other if [ "$vox_size" -lt "$hires_voxsize_threshold" ]

LeHenschel: (i.e. instead of 
line 637:  cm="-noconform" else cm=""
line 727: hsuff=".predec" else hsuff="")

recon_surf/recon-surf.sh
```diff
@@ -753,7 +744,7 @@ else
     RunIt "$cmd" $LF $CMDF
     
     # for hires decimate mesh 
-    if [ "$vox_size" == "auto" ]; then
+    if [ "$vox_size" -lt "$hires_voxsize_threshold" ]; then
```
LeHenschel: just check if hires is set, no need for the -lt.

FastSurferCNN/run_prediction.py
```diff
@@ -169,10 +171,10 @@ def conform_and_save_orig(self, orig_str: str) -> Tuple[nib.analyze.SpatialImage
         # Save input image to standard location
         self.save_img(self.input_img_name, orig_data, orig)
 
-        conform_min = self.vox_size == "auto"
-        if not conf.is_conform(orig, conform_min=conform_min, check_dtype=True, verbose=False):
+        if not conf.is_conform(orig, conform_vox_size=self.vox_size, check_dtype=True, verbose=False,
+                               conform_to_1_threshold=self.conform_to_1_threshold):
```
LeHenschel: I do not find this flag very descriptive. Would rather set it to conform_to_one_mm or _1mm, but I guess that is personal preference

LeHenschel (commented):
recon-surf.sh can be simplified even further in my opinion. I do not like the conform_to_1_threshold name because it implies you are conforming to (any) one resolution and not to 1.0 mm.

FastSurferCNN/run_prediction.py
dkuegler: we can change that to conform_to_1mm_threshold

FastSurferCNN/data_loader/data_utils.py
dkuegler: I just changed the call interface to work as before.
`load_and_conform_image()` is actually a stub as it is not called anywhere else as far as I can see.

LeHenschel: Yeah, but if someone is using it in the future, would it not make sense to either remove the function or adopt it to work in the same way (i.e. pass the voxsize directly to the conform call)?

recon_surf/recon-surf.sh
dkuegler: While in the following code, we could also test against `$hiresflag`, here we need to SET the hiresflag variable.
Also, I would want to have a string in the log somewhere that documents which mode was selected (although that would also always be apparent by the existence of `-hires` in later calls. Still having this here and also the `$hires_voxsize_threshold` documented seems like a good idea to me.

I'll push a patch to clean up the ifs.

recon_surf/recon-surf.sh
dkuegler: I think it is better to do this here, so you can change the threshold in recon-surf. Even a parameter/config setting later.

recon_surf/recon-surf.sh
dkuegler: 1. It checks conformity between the two.
2. We don't need to do a bunch of error checking if we got the right output in the first call (i.e., if the output of call # 1 or the cut-commands were unclear, the second call will fail)

recon_surf/recon-surf.sh
dkuegler: Personally, I would clear statements and `[ ! -z "$hiresflag" ]` does not really seem "clearer" to me, i.e. I know have to look up what empty `$hiresflag` means...

dkuegler:
Can also be added later, but right now, we cannot change the --conform_to_1mm_threshold in `run_fastsurfer.sh`. `run_fastsurfer.sh` fixes this to 0.95 right now. In my opinion, this is fine since users can still use the `--vox_size` flag to force a specific behavior, but this is a limitation currently.

recon_surf/recon-surf.sh
LeHenschel: I think hiresflag is very descriptive = hires processing (voxsize as well), whereas voxsize_threshold is unclear.

recon_surf/recon-surf.sh
LeHenschel: The voxsize_threshold is set to 0.999. This can not be changed. So in my opinion this just complicates recon-surf unnecessarily. Also, either the input is 1 or below 1 = hires processing. Having variable voxelsizes already does not make sense here. I would just have a hires flag and turn this on if the voxsize flag in run_fastsurfer.sh is set accordingly. Then have a check in conform.py --check_only if the hires flag is passed and resolution is below 1.0 mm and abort if this is not the case. The voxsize threshold and voxsize flag only makes sense if you have a variable hires_voxsize_threshold to run 1.0 mm settings with hires images, but this is not tested (neither for FastSurfer nor FreeSurfer). So do we really want to support this?

Alternatively: Pass the voxsize (always a float number), then just check in conform if the resolution of the image == voxsize (for the image and the segmentation) and abort if this is not the case. In recon-surf.sh voxsize is then a necessary parameter that has to be passed (the automatic option still exists if you run from run_fastsurfer.sh)

PR #233, opened 2023-01-10
Fix: Padding size type

af-a:
## Description

This PR makes a small modification in the behaviour of the `ZeroPad2D` and `ZeroPad2DTest` objects in `augmentation.py`: both now expect integers instead of floats in the specified tensor sizes passes in as inputs.

This avoids an error when concatenating the tensor sizes with the input image's number of channels returned by `numpy`, which is an integer, in the following line: https://github.com/Deep-MI/FastSurfer/blob/55d20b831b3a8349e49e81229b7a0a0381bf55e1/FastSurferCNN/data_loader/augmentation.py#L121

If the input tuple to `np.zeros()` has a combination of integers and floats, it raises the following error, which is avoided with this fix:
```
TypeError: 'float' object cannot be interpreted as an integer
```

### DIFF
--- a/FastSurferCNN/data_loader/augmentation.py
+++ b/FastSurferCNN/data_loader/augmentation.py
@@ -15,6 +15,8 @@
 
 
 # IMPORTS
+from numbers import Number
+
 import numpy as np
 import torch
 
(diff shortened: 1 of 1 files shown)


FastSurferCNN/data_loader/augmentation.py
```diff
@@ -107,7 +107,7 @@ def __init__(self, output_size, pos='top_left'):
         :param output_size:
         :param pos: position to put the input
         """
-        if isinstance(output_size, float):
+        if isinstance(output_size, int):
```
dkuegler: probably should be `isinstance(output_size, numbers.Number)`, and then `output_size = (int(output_size),) * 2`
same for ZeroPad2DTest

af-a: Done (see below).

PR #241, opened 2023-01-23
Surfreg script

agirodi:
This PR adds:
- A new script "recon-surfreg.sh" that solely does the surface registration
  - Having similar behaviour/flags as recon-surf.sh
- minor typo fixes

### DIFF
--- a/recon_surf/recon-surfreg.sh
+++ b/recon_surf/recon-surfreg.sh
@@ -360,12 +360,6 @@ then
   exit 1
 fi
 
-# Check if running on an existing subject directory
-if [ -f "$SUBJECTS_DIR/$subject/mri/wm.mgz" ] || [ -f "$SUBJECTS_DIR/$subject/mri/aparc.DKTatlas+aseg.orig.mgz" ]; then
-  echo "ERROR: running on top of an existing subject directory!"
-  echo "The output directory must not contain data from a previous invocation of recon-surf."
-  exit 1
-fi
 
 # Check input segmentation quality
 echo "Checking Input Segmentation Quality ..."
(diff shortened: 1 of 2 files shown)


recon_surf/recon-surfreg.sh
```diff
@@ -0,0 +1,499 @@
...
+  CMDF="$SUBJECTS_DIR/$subject/scripts/$hemi.processing.cmdf"
+  CMDFS="$CMDFS $CMDF"
+  rm -rf $CMDF
+
+  echo "echo \" \"" |& tee -a $CMDF
+  echo "echo \"============ Creating surfaces $hemi - FS sphere, surfreg ===============\"" |& tee -a $CMDF
+  echo "echo \" \"" |& tee -a $CMDF
+
+  # Surface registration for cross-subject correspondance (registration to fsaverage)
+    cmd="recon-all -subject $subject -hemi $hemi -sphere $hiresflag -no-isrunning $fsthreads"
```
m-reuter: This just calls mris_sphere , see https://github.com/freesurfer/freesurfer/blob/483c25eef64b6352a93e950b9ad31a4d90117e10/scripts/recon-all#L3875 
for that $hiresflag is not needed, so you can remove it here and the block above that checks voxel sizes.

agirodi: The same command is used in the [recon-surf.sh](https://github.com/Deep-MI/FastSurfer/blob/3c84f9d8e36a4836d26a6a47919c9d315b83093b/recon_surf/recon-surf.sh#L825). Since the `$hires` flag is not needed for `mris_sphere`, it can be removed there as well?

m-reuter: in recon-surf the highresflag is passed to all recon-all calls because of two reasons: 

- no time to check which individual block really use it, sometimes we call multiple at a time
- to be future proof, in case that high-res functionality is added in the future to a specific block.

So I would, for now, pass it everywhere in recon-surf, since it is populated anyway. 
Here I think we can go either way, since it is a single command we may even want to directly call the mris_sphere command instead of using the recon-all as a wrapper.

PR #245, opened 2023-01-30
Fix: Incorrect `mri_segstats` argument

af-a:
## Description

This PR fixes a bug in an `mri_segstats` call in `recon-surf.sh` ([this line](https://github.com/Deep-MI/FastSurfer/blob/617ed0d87372eafe94683bd7b2e1096a2727966e/recon_surf/recon-surf.sh#L660)), in which an incorrect argument was specified (`segfile` instead of `seg`).

This was pointed out by a user who tried running the pipeline with the `vol_segstats` flag, which leads to the error:
```
ERROR: Option --segfile unknown
setting seed for random number genererator to 1234
Command exited with non-zero status 255
```


af-a:
The changes introduced here have been transferred to https://github.com/Deep-MI/FastSurfer/pull/246

PR #248, opened 2023-02-03
fix crash on native install on windows

dkuegler:
When installing Fastsurfer natively on windows, the script to download the checkpoints crashes because the url is malformed.

This is specifically, because we are using os.path.join to create the url, but on windows, os.path.join uses backslash to form the path. Http needs forward slashes, so we should just fall back to string concatenation here 

This PR still needs to be tested, 1. On Linux and 2. On windows.


dkuegler:
I guess I have to rebase the commit...

dkuegler:
Tested and already pushed to dev

PR #255, opened 2023-02-14
Performance optimization and refactoring of the core prediction loop

dkuegler:
Refactoring and simplification of the reading of subjects for batch processing. Also, some parallelization of IO of the pipeline, which is most likely to affect the speed of batch processing (single subject processing is less affected, because it cannot be pipelined).

New speed for N=10: 329.0 s
Old speed for N=10: 387.2 s

Note: Pipelining and asynchronous execution (specifically for IO, which becomes apparent in the log) can be deactivated with --threads 1.

Changelist:
FastSurferCNN
- data_loader
  - conform: - update doc: Add Notes to some functions that work based on header information alone
  - dataset:
    - remove argument img_filename, which is only moved for the logger (which produces highly redundant log messages).
- util
  - __init__: - add common as a module
  - common:
    - add SubjectDirectory and SubjectList classes which manage the file names and paths to filenames in the subject directory (which will also be used in CerebNet for a consistent commandline interface).
    - SubjectList creates the list of subjects from a csv file, a search pattern or an absolute path. - SubjectList / SubjectDirectory include improved error messages and descriptions as well as more streamlined, consistent and error checked parameter processing. - move the find_device function to common for consistency reasons (misc is only used in training, common is only used in inference).
  - misc: - move the find_device function to common for consistency reasons (misc is only used in training, common is only used in inference).
  - parser_defaults:
    - add option to "overwrite" default arguments, e.g. help text, dest and default value by passing overwrite.
    - added documentation and examples - add threads flag
- inference
  - run: rename argument img_filename to image_name, where it is only used as a verbosity output.
- run_prediction
  - add some parallelization of reading (and saving) of files
  - added simple pipelining of the loading and conforming step, so process never waits for IO
  - consistency of the interface
  - rename argument img_filename to image_name, where it is only used as a verbosity output.
  - change script to use the SubjectList and SubjectDirectory interface
  - if we are running with ab absolute path, but we cannot say where to store the aseg or the brain_mask (because of missing flags), so not calculate those images.
- segstats
  - change the call interface to write_statsfile -> now does not use the args object any more.

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -261,6 +261,9 @@ def find_min_size(img: nib.analyze.SpatialImage, max_size: float = 1) -> float:
 
     Returns:
         The rounded minimal voxel size
+
+    Note:
+        This function only needs the header (not the data).
     """
     # find minimal voxel side length
     sizes = np.array(img.header.get_zooms()[:3])
@@ -281,6 +284,9 @@ def find_img_size_by_fov(img: nib.analyze.SpatialImage, vox_size: float, min_dim
 
     Returns:
         The number of voxels needed to cover field of view.
+
+    Note:
+        This function only needs the header (not the data).
     """
     if vox_size == 1.:
         return min_dim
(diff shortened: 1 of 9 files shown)


dkuegler:
I am not really sure, if we want to activate the pipelining and asynchronous IO by default, if we do not, we can set the default value for --threads to 1, e.g. by
`parser_defaults.ALL_FLAGS['threads'](parser, default=1)` instead of putting threads in the list in https://github.com/Deep-MI/FastSurfer/blob/53433729f36ac97a160d803e5e17edd3979472fd/FastSurferCNN/run_prediction.py#L279

We also probably want to fix the interaction of the logger with tqdm...

PR #257, opened 2023-02-15
Bugfix: Wrong number of parameters for "setup_logging"

ClePol:
The function setup_logging expects one parameter, but two are given in train.py

### DIFF
--- a/FastSurferCNN/train.py
+++ b/FastSurferCNN/train.py
@@ -47,7 +47,7 @@ class Trainer:
 
         # Create the checkpoint dir.
         self.checkpoint_dir = cp.create_checkpoint_dir(cfg.LOG_DIR, cfg.EXPR_NUM)
-        logging.setup_logging(cfg.LOG_DIR)
+        logging.setup_logging(cfg.LOG_DIR, cfg.EXPR_NUM)
         logger.info("Training with config:")
         logger.info(pprint.pformat(cfg))
         self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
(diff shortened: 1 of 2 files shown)


ClePol:
This bug also affects the stable branch

LeHenschel:
Thanks for reporting this. Removing the argument in train.py does solve the error message, but then the log-file will not be set up correctly. In this case the cfg.LOG_DIR will be treated as the log-file and the command fails as well (because it is a directory, not a file). I would instead propose some changes to the logging function setup_logging @m-reuter

ClePol:
I encountered this behavior as well - I think this can be circumvented by either specifying logfiles (instead of directories) in the .yml or modifying the setup_logging fucntion to work with directories and filenames. In my local copy I made a modification that appends the file ending 'log.txt' if the input is a directory.

If you agree, that this is good default behavior I can also add these modifications here.

ClePol:
After looking more closely at your commit this is basically what you did as well. So I think that's a good solution.

LeHenschel:
In the .yaml file there is only a LOG_DIR variable and this needs to be a directory as it is also used to set up the checkpoint, config, summary directories and so on. Appending log.txt is one option. With my commit, it is using the Experiment number as the filename (expr_${expr_num}.log). Can be changed to preference.

ClePol:
I think changing back to two parameters might break the inference
https://github.com/ClePol/FastSurfer/blob/stable/FastSurferCNN/run_prediction.py#L337

I patched together a case handling that adresses both experiment number and directiory during training and then the full logfile path during inference - but this should probably be two different functions..

LeHenschel:
We can also just roll back and join LOG DIR and EXPR_NUM in train.py and use the old setup_logging function. That is the easiest fix and given that david has multiple changes in his pull request the easiest for now.

So just replacing line 50 in train.py with:

```
logging.setup_logging(os.path.join(cfg.LOG_DIR, "logs", cfg.EXPR_NUM + ".log"))
```

PR #263, opened 2023-02-21
Remove np.float, deprecated in numpy 1.20 and removed in 1.24

mscheltienne:
I'm going to assume that changes have to be merged in the `dev` branch.
The alias `np.float` has been deprecated a long time ago, and is now removed from recent version of `numpy`.


m-reuter:
Thanks, yes, this is on our list already. Your help is appreciated!

PR #274, opened 2023-03-08
Fixed bug in FastSurferVINN interpolation

ClePol:
Fix proposed by @dkuegler 
Also affects stable branch

This bug causes wrong rescale factors to be computed, which can crash training (and affect training results?)


m-reuter:
Does it also affect inference, or only training?

ClePol:
No, as far as I understand this only affects batches with different resolutions in the batch - which should not happen during inference.

dkuegler:
> No, as far as I understand this only affects batches with different resolutions in the batch - which should not happen during inference.

Yep. Only training like @ClePol mentioned.

LeHenschel:
Also just to clarify, this version was not used for the VINN training. The changes were added for the GPU memory optimization (https://github.com/Deep-MI/FastSurfer/pull/176). For inference this did not make any differences (checked on 80 subjects).

PR #276, opened 2023-03-10
move segstats into segmentation pipeline

dkuegler:
This PR introduces the segstats.py script capabilities into the FastSurfer pipeline for CerebNet and FastSurfer (volseg).
The point is that we want to produce PV-corrected statsfiles for CerebNet as part of the CerebNet segmentation pipeline.
This requires the following sigfinicant changes:
1. perform bias field correction in the segmentation pipeline (instead of surface)
2. unclear: what should change in recon-surf.sh
3. unclear: move volsegstats (statsfiles for segmentation results only, no surface updates) into the segmentation pipeline
4. unclear: based on 3., the corpus callosum is not segmented at that point, so do we perform the segstats without CC here?

fix call to segstats in CerebNet/inference.py and CerebNet/run_prediction.py

This PR is currently WIP given the intended behavior is not defined yet.


dkuegler:
At this point the script is ready for review.
In particular, a CerebNet statsfile is automatically generated and a file can be generated for aparc+aseg (use `--vol_segstats` flag in `run_fastsurfer.sh`, which at this point generates both the segstats of the new style and the old style).

This version is probably not final, but it needs to be reviewed by people familiar with FreeSurfer-compatiblity and expectations in mind.

dkuegler:
A lot of the filenames and conventions might not make perfect sense, also etiv and others are missing from aparc segstats. @m-reuter

kdiers:
Here's a comparison of a Cerebellum.cerebNet.stats file against a standard aseg.stats file from FS7.1:

### What is missing in the Cerebellum.cerebNet.stats file:
- \# SUBJECTS_DIR: --sd is a required argument for run_fastsurfer.sh (and inferred in run_prediction.py if running this file directly?), so could in principle be included in the stats file?
- \# subjectname: --sid is a required argument for run_fastsurfer.sh (and inferred in run_prediction.py  if running this file directly?), so could in principle be included in the stats file?
- \# Measure \<measure\>: no measures are included in the Cerebellum.cerebNet.stats file. And since all of them refer to some whole-brain- or cortex-related property, I think this is the correct decision. You could invent new ones like "CerebVol" etc., but I do not see a strong need.
- \# PVVolFile: not sure if this applies at all to the  Cerebellum.cerebNet.stats file?
- \# <SegVol|ColorTable|PVVolFile>Timestamp: can probably be safely ignored
- \# <Excluding|ExcludeSegId>: does not apply to cerebellum segmentation in my eyes, so can probably be left out.

### What is not included in the aseg.stats file (but in Cerebellum.cerebNet.stats):
- \# Only reporting non-empty segmentations: comments like this can probably safely be added in the header section

### Other differences:
- \# TableCol: index start with 0 in the Cerebellum.cerebNet.stats file, and with 1 in the aseg.stats file
- an "Index" entry is missing from the "# ColHeaders" line; "Index" should be the label of the first column
- NTableCols should be 10 (not 9) in Cerebellum.cerebNet.stats
- Mismatch between numeric label IDs: e.g., 7/46 are Left/Right_Corpus_Medullare in  Cerebellum.cerebNet.stats, but Left/Right-Cerebellum-White-Matter in FreeSurferColorLUT.txt
- Mismatch between numeric label IDs: e.g., 601 is "Cbm_Left_I_IV" in FreeSurferColorLUT.txt, but "Left_I_IV" in (applies to other 6.. labels as well)

kdiers:
Here's a comparison of an aparc.nosurf.stats file against a standard aseg.stats file from FS7.3.2:

See the [Morphometry stats](https://surfer.nmr.mgh.harvard.edu/fswiki/MorphometryStats) page in the FreeSurfer wiki for information how the summary measures in the header section are calculated.

### What is missing in the aparc.nosurf.stats file:

- \# SUBJECTS_DIR: --sd is a required argument for run_fastsurfer.sh (and inferred in run_prediction.py if running this file directly?), so could in principle be included in the stats file?
- \# subjectname: --sid is a required argument for run_fastsurfer.sh (and inferred in run_prediction.py if running this file directly?), so could in principle be included in the stats file?
- \# Measure \<BrainSeg|BrainSegNotVent|VentricleChoroidVol|SubCortGray|TotalGray|Mask\>: all these are volume- / voxel-based estimates and could be computed as far as I can see. See the [Morphometry stats](https://surfer.nmr.mgh.harvard.edu/fswiki/MorphometryStats) page for details.
- \# Measure \<lhCortex|rhCortex|Cortex\>: involve surface-based volume estimates; is *approximately* equal to the count of cortical label voxels in aparc+aseg.mgz
- \# Measure \<lhCerebralWhiteMatter|rhCerebralWhiteMatter|CerebralWhiteMatter\>: involve surface-based volume estimates; is *approximately* equal to the count of WM voxels in aparc+aseg.mgz
- \# Measure \<SupraTentorial|SupraTentorialNotVent\>: involve surface-based volume estimates; probably not easy to include
- \# Measure \<BrainSegVol-to-eTIV|MaskVol-to-eTIV|EstimatedTotalIntraCranialVol\>: requires alternative eTIV calculation, so can only be included after alternative eTIV calculation is available
- \# Measure \<lhSurfaceHoles|rhSurfaceHoles|SurfaceHoles\>: require surfaces / topology fixing and cannot be included unless surface stream is run
- \# PVVolFile: could include this to indicate which file the partial volume calculation is based upon (also include PVVolFileTimeStamp)

### What is not included in the aseg.stats file (but in aparc.nosurf.stats):

- \# Only reporting non-empty segmentations: comments like this can probably safely be added in the header section

### Other differences:

- an "Index" entry is missing from the "# ColHeaders" line; "Index" should be the label of the first column
NTableCols should be 10 (not 9) in aparc.nosurf.stats
- change timestamp format to YYYY/MM/DD HH:MM:SS
- naming: at first sight, one might expect similarity between the aparc.nosurf.stats file and FreeSurfer's [lr]h.aparc.stats files; but I think that this is misleading since there are no surface-based measures in aparc.nosurf.stats (kind of contradiction in itself). In fact, I believe we are rather comparing to an aseg.stats file with added volume-based aparc estimates. However, there is no aparc+aseg.stats file in FreeSurfer (although it could be, in principle), so I guess the aseg.stats file is the closest corresponding file.

kdiers (commented):
Just to close the request for review. See above comments for details.

dkuegler:
Some of these items are easy to fix/include, some are more difficult:

- SUBJECTS_DIR, subjectname, *Timestamp added
- Timestamp format changed
- ExcludeId renamed to ExcludeSegId, Excluding is ignored (
- Index: added
- default naming changed to aseg+DKT.stats
- The doubling of InVolFile and PVolFile does not really make sense, to me, but I am putting it in there fore now.
- Label name mismatches for CerebNet, I had kept the label names of the CerebNet labels from the CerebNet lut, that explains that discrepancy.... But it should really be FreeSurfer/FastSurfer names and SegIds (the latter is already the case....)

PR #278, opened 2023-03-20
add dockerhub deployment workflow 

agirodi:
This PR adds a GitHub Actions workflow, for building all the Docker images and pushing them to the Dockerhub. 
* Triggered by a manual dispatch (Can be changed if desired)
* Steps:
  * Checkout and fetching repo
  * Logins to Dockerhub
  * Gets the latest used Tag
  * Builds and pushes all images for every Dockerfile (Except experimental AMD)
    * uses the latest tag for the docker version as in `deepmit/fastsurfer:gpu-[TAG]`. So consistent tagging is needed.

#
This should work, but I would recommend, first getting a local runner, as it put out a warning last time about lack of disk space:
 > You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 82 MB

The workflow then needs to be updated to use the new runner.

Also, there needs to be a GitHub [secret](https://github.com/Deep-MI/FastSurfer/settings/secrets/actions) for the username and token in able for this to work.

To check the latest functioning run of this Workflow, see [here](https://github.com/agirodi/FastSurfer/actions/runs/4467717898)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,9 @@
 name: deploy-docker
 
-on: [workflow_dispatch]
+on:
+  release:
+    types:
+      - published
 
 jobs:
   deploy-gpu:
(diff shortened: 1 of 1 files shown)


m-reuter:
This probably needs updating after pr #215

m-reuter:
@agirodi can you update this to our new build scripts (see docker directory). Also we only build a cpu and a gpu (cuda) version at the moment as surf-only and segmentation-only can be called from those images.

dkuegler:
Generally, this should just be `PYTHONPATH={Fastsurferdir} python Docker/build.py --device (cuda|cpu) --tag deepmi/fastsurfer:gpu-v{version}` or similar. Ask if there are questions

agirodi:
The workflow has now been changed to work with the new build script. However, with the way workflows are handeled, even if this is merged into the dev branch it can not be run or used until the main branch has at least an empty `deploy.yml` file. I did that in my fork and the last deploy run can be checked out [here](https://github.com/agirodi/FastSurfer/actions/runs/7211351462). Also the repository need a docker username and token in the secrets to be added.

m-reuter:
Thanks! Please update, so that it triggers automatically at a release (published) in stable branch.

Also update these additional tags for published releases on the stable branch:
- latest (point to gpu)
- gpu-latest
- cpu-latest

For manually triggered builds on the dev branch, we could use a different tag, such as dev-gitcode. But that functionality is not so important, as we build dev locally. Probably easier and safer to remove the manual trigger completely instead.

Issue #281, opened 2023-04-02
freesurfer qcache error after running fastsurfer - no jacobian_white


pavgreen9:
## Question/Support Request

Hi, I ran fastsurfer just fine but require the qcache fwhm output for mris_preproc analysis. A previous user had a [similar issue](https://github.com/Deep-MI/FastSurfer/issues/49) and fixed it by manually inputting the --surfreg flag and their qcache output turned out fine. 

I've noticed that the --surfreg flag has been automatically enabled since then given rh/lh.sphere that is already in my surf directory. However, when I run qcache in recon-all I get the error 

```
ERROR: cannot find /home/pavgreen/Documents/LRGS/MRI/Participants/fast_out/A686/surf/lh.jacobian_white
Command exited with non-zero status 1
```

## Environment
 - FastSurfer Version: 2.0.4 Docker GPU
 - FreeSurfer Version: 7.3.2 
 - OS: Ubuntu 22.04.2
 - CPU: Ryzen 5 3600
 - GPU: GTX 1070 Ti

### Execution

**Fastsurfer**: 
while read participant; do
	nohup docker run --gpus all \
		-v /home/pavgreen/Documents/LRGS/MRI/Participants/unprocessed_raws/${participant}/mprage/anat:/data \
		-v /home/pavgreen/Documents/LRGS/MRI/Participants/fast_out:/output \
		-v /usr/local/freesurfer/7.3.2:/fs_license \
		--rm --user $(id -u):$(id -g) deepmi/fastsurfer:latest \
		--fs_license /fs_license/license.txt \
		--t1 /data/${participant}.nii \
		--sid ${participant} --sd /output \
		--parallel \
		--threads 12 \
		--surfreg \
		--vox_size 1
	sleep 90S
done < subjList.txt

**Freesurfer qcache**:
recon-all -s A686 -qcache

Thank you

LeHenschel:
Seems like the jacobian is not generated by default in FastSurfer. You can create it with  FreeSurfer:

mris_jacobian ?h.white ?h.sphere.reg ?h.jacobian_white

We might want to add this by default @m-reuter @AhmedFaisal95 . The command takes less than a second.

m-reuter:
Yes, should be done at same spot as in recon-all. Not sure how this got lost. Is anything else missing?

LeHenschel:
[...]

PR #282, opened 2023-04-04
Added call to create jacobian_white measure directly in fastsurfer

LeHenschel:
Previously jacobian_white was not created in FastSurfer (as pointed out by @pavgreen9 in #281 ). If --surf-reg was run (on by default) this can be easily generated in less than a second. Command is added after white and pial are generated (grouped with other surface measures e.g. curvature, area and thickness).


LeHenschel:
Tested on two subjects (1.0 mm and 0.8 mm) with and without surf-reg (jacobian will only be created if for surf-reg).

PR #283, opened 2023-04-04
Added missing allow_root flag to python calls run_prediction and segstats

LeHenschel:
Runs executed as root failed even when --allow_root was passed due to missing passing of the arguments in run_fastsurfer.sh to the python modules. This pull requests fixes the issue.


LeHenschel:
Checked on two subjects (0.8 and 1.0 mm) and ran through successfully.

PR #291, opened 2023-05-05
 segstats FreeSurfer legacy

dkuegler:
Deactivate the freesurfer legacy implementation by default and add a flag to enable it.

Still needs to be properly tested before merge.


dkuegler:
Documentation and default behavior changes

dkuegler:
Testing confirmed segstats.py was not broken, but norm.mgz and orig_nu.mgz are different in that one is masked by the brainmask, which affects the mixing coefficient of mixing between background and regions bordering the background, because the average intensity of the background changes (masked background voxels).

Issue #280, opened 2023-03-23
ipynb tutorial: "Run Interact" is not working. While clicking the button only white space is displayed.


shijinknox007:
Hi all,

I am trying to execute Complete_FastSurfer_Tutorial.ipynb in colab. Then I came across with some issues.

1. I am using "Alternative" option to upload the input image (tutorial image)
2. Run fastsurfer
3. Plot some slices from the segmented brain - then I got an error "ValueError: keyword grid_b is not recognized" (but Slices, Prediction, and Overlay_GT images were displayed)
4. Upon clicking "Run Interact" - Nothing is displaying but a white space

Thanks and Regards,
Shijin Knox G.U

LeHenschel:
The Value Error appears because the first parameter of Axes.grid and Axis.grid has been renamed to visible (was previously named b). Can be fixed changing ax.grid(b=None) to ax.grid(visible=None) in the corresponding function for slice plotting. The new matplotlib version might also cause the bug of not correctly displaying the figure for the mesh plot. @agirodi @AhmedFaisal95 please have a look.

agirodi:
The ValueError was fixed by `b=None` to `visible=None`.

For the second error, I can only say that matplotlib is not the cause of the error. It neither seems to be used in that code cell nor did downgrading the matplotlib to the previous version fixes the error. I tried downgrading some other packages, even forcing the entire Fastsurfer dependencies versions, to no result.
According to the [release notes](https://colab.research.google.com/notebooks/relnotes.ipynb), these are the changed packages we use, since the last NB update (as far as I can tell). Its probably one of those:

> absl-py
> cachetools
> charset-normalizer
> importlib-metadata
> matplotlib
> networkx
> numpy
> packaging
> pandas
> pillow
> protobuf
> requests
> scikit-image
> scikit-learn
> scipy
> tenacity
> tensorboard
> tifffile
> torch
> torchvision
> typing-extensions
> urllib3
> werkzeug
> wrapt
> zipp

I can look into which one exactly does the error, by downgrading each. But maybe someone can already see where the problem may lie.

shijinknox007:
> The ValueError was fixed by `b=None` to `visible=None`.
> 
[...]

PR #300, opened 2023-05-16
Fix: 3D interact showing white screen

agirodi:
This PR fixes the 3D interact, that has shown a white screen/no image before.

The bug was discussed in #280 and occurred because GoogleCollab upgraded its preinstalled packages.
The only change is to downgrade the **plotly** version back to 5.9.0 (was 5.13.1), the same version that is required by FastSurfer anyways.

An alternative fix is possible, where the requirements.txt is installed instead, which could make this more stable for future changes to our FastSurfer dependencies and Collabs changes. But it takes several minutes to install and I recommend doing this only if necessary.


LeHenschel:
Looks good. I agree to keep it as is for now. In the future, we should check what changed between the plotly versions to see if we can adapt the plotting script itself.

PR #301, opened 2023-05-16
Fixed bug in generate_hdf5.py

larshaupt:
Bug was that zoom=zoom[1:] (and similar operations) was executed multiple times such that zoom would end up being empty []


LeHenschel:
Zooms should not be empty, even if it is inside the for loop as it is taken from the original variable zoom and not indexed with i. Potentially, you had an issue with stable, but in dev everything should be working fine after #251

Issue #290, opened 2023-05-04
Manual Editing Brain Mask


psimonjones:
Is there a way to add your own brainmask and use the -noskullstrip option as can be done in FreeSurfer recon? FastSurfer if chopping off the temporal lobes of atrophied brains in a similar way that Freesurfer does, but I cannot see how you fix these issues.

m-reuter:
Hi, the brainmask gets created from the segmentation (some spatially varying iterations of dilation and erosion of the union of all labels). That means you can edit the brainmask and then run the recon-surf part which will automatically use the modified brainmask. It probably make sense to also edit the segmentation directly and create the correct gray matter and white matter labels in that region. The recon-surf part should then work as expected. 

Do you have example cases (e.g. IDs from publicly available datasets where you see this problem)? We probably can look for it in the training data and eventually fix it, so that FastSurfer does not replicate those issues in a future version.

psimonjones:
Using Freesurfer pipeline we might typically have to edit/provide brainmask.mgz, colour in bits of aseg.presurf.mgz and also add white matter control points. Following your comments and looking at the time stamps of the output of run_fastsurfer.sh I see the brainmask.mgz is downstream of some other files like aparc.DKTatlas+aseg.deep.mgz, mask.mgz, aparc.DKTatlas+aseg.orig.mgz. Do you mean edit the file aparc.DKTatlas+aseg.deep.mgz? I may be able to deface an example to send to you.

m-reuter:
Yes the pipeline in FastSurfer works different from FreeSurfer. The order of things is slightly different. 
Our segmentation module is at the beginning and creates the aparc.DKTatlas+aseg.deep.mgz based on which it then creates the brainmask and also reduces the labels to an aseg.mgz. After manual edits to the aparc.DKT... this can be repeated via the command:
`python3 reduce_to_aseg.py -i aparc.DKTatlas+aseg.deep.edited.mgz -o subject/mri/aseg.auto_noCCseg.mgz -outmask subject/mri/mask.mgz --fixwm`
After that you can start the surface module (recon-surf). Sometimes it may be sufficient to only edit the brainmask and restart the surface pipeline but if GM and WM regions are missing, you probably need to add them back in in the segmentation file first. 

[...]

PR #304, opened 2023-05-20
Create EDITING.md

m-reuter:
I am creating a file describing the possible edits in FastSurfer. (see also #290 ). 
It would be important that these edits get tested before we merge this. 
It may be necessary to adjust the description or even the code. For example at some point we stopped when some files already existed in the subject directories. We need to make sure that this does not prevent the edits.


m-reuter:
closes #290 for now

PR #310, opened 2023-05-24
Add additional code to make sure segstats is correct

dkuegler:
see commits
mostly minor edits
F.e. by default put 3 digits after the dot for Volumes
big edits improve and document the differences between freesurfer and segstats.py's freesurfer_legacy mode.

@m-reuter I reverted the code formatting changes, you need to rerun that if you want those changes.


m-reuter:
Sadly it is now almost impossible to see your edits. It would have been better to resolve conflicts and (if even necessary) revert formatting in a different PR. I will merge this anyway in the interest of time.

PR #322, opened 2023-05-31
small edits

greydongilmore:


m-reuter:
hi, thanks for contributing. PRs should only go into dev branch. Stable is for releases. Also python versions and package versions are tested and cannot simply be changed without our thorough release testing. 
When using versions outside our specifications, results can change or things can break (even though in most cases it would just work).

PR #325, opened 2023-06-01
hotfix: --no_fs_t1 lower case

### LINKED ISSUE #324: no_fs_T1 not passed correctly between run_fastsurfer and recon-surf
When using --no_fs_T1 in run_fastsrufer one gets this error:
```
./recon-surf.sh --sid sub-01_ses-01 --sd /out/fastsurfer --t1 /out/fastsurfer/sub-01_ses-01/mri/orig.mgz --asegdkt_segfile /out/fastsurfer/sub-HC062_ses-01/mri/aparc.DKTatlas+aseg.deep.mgz     --parallel --threads 15 --py python3.8  --no_fs_T1 --allow_root
ERROR: Flag --no_fs_t1 unrecognized.

```

Inside `recon-surf.sh` the code looks like this:
```
    --no_fs_T1 )
    get_t1=0
    shift # past argument
    ;;
```

I removed the flag `--no_fs_T1` from my code and it worked, maybe updating to `--no_fs_t1` in `recon-surf.sh` could solve this issue.

_Originally posted by @rcruces in https://github.com/Deep-MI/FastSurfer/issues/319#issuecomment-1571082463_

m-reuter:
closes #324

### DIFF
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -300,7 +300,7 @@ case $key in
     check_version=0
     shift # past argument
     ;;
-    --no_fs_T1 )
+    --no_fs_t1 )
     get_t1=0
     shift # past argument
     ;;


LeHenschel:
So if you want to do the fix this way around (changing it in recon_surf.sh), the documentation needs to be updated as well. Both in recon_surf.sh and run_fastsurfer.sh the flag is described with upper case. Otherwise, it would have been easier to change it in run_fastsurfer.sh because this is the only place where the T was lower case.

m-reuter:
I thought, when we check for parameters, we always convert the input to lower case and then compare the strings in lower case format. That is why I changed it like this.

LeHenschel:
Ok, I see. Then it makes sense. I was not aware of the new changes there.

PR #327, opened 2023-06-02
Add version info for run_fastsurfer and into the docker image

dkuegler:
- Adds extra option to run_fastsurfer.sh: --version with extra option long
  - outputs a version identifier (`{VERSION_TAG}-{git-hash} ({git-branch})`)
  - long option (`--version long`) also adds md5sums of checkpoints and git status to document difference to most recent git commit
- Add VERSION file to docker images
- add version to logfile

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = 'setuptools.build_meta'
 
 [project]
 name = 'fastsurfer'
-version = "2.1.1"
+version = "2.2.0dev"
 description = 'a fast and accurate deep-learning based neuroimaging pipeline'
 readme = 'README.md'
 license = {file = 'LICENSE'}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,65 @@
+[build-system]
+requires = ['setuptools >= 61.0.0']
+build-backend = 'setuptools.build_meta'
+
+[project]
+name = 'fastsurfer'
+version = "2.1.1"
+description = 'a fast and accurate deep-learning based neuroimaging pipeline'
+readme = 'README.md'
+license = {file = 'LICENSE'}
+requires-python = '>=3.8'
+authors = [
+    {name = 'Martin Reuter', email = 'martin.reuter@dzne.de'},
+    {name = 'Leonie Henschel', email = 'leonie.henschel@dzne.de'},
+    {name = 'David Kügler', email = 'david.kuegler@dzne.de'},
+]
+maintainers = [
+    {name = 'Martin Reuter', email = 'martin.reuter@dzne.de'},
+    {name = 'Leonie Henschel', email = 'leonie.henschel@dzne.de'},
+    {name = 'David Kügler', email = 'david.kuegler@dzne.de'},
+]
+keywords = [
+    'python',
+    'Deep learning',
+    'Segmentation',
+    'Brain segmentation',
+    'Brain analysis',
+    'volumetry',
+]
+classifiers = [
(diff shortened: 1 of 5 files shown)


m-reuter:
thanks, but version tags should be read from GitHub (or in the future from the toml file), so that they are only in one place.

dkuegler:
sure, and all but the "VERSION_TAG" is read from git, but right now, we do not support this in our git.

run_fastsurfer.sh
```diff
@@ -247,20 +247,29 @@ EOF
...
+    pushd  "$FASTSURFER_HOME" > /dev/null || return
     VERSION_INFO="$VERSION_TAG-$(git rev-parse --short HEAD) ($(git branch --show-current))"
+    popd > /dev/null || return
+  else
+    VERSION_INFO="$([ -f "$VERSION_FILE" ] && head "$VERSION_FILE" -n 1 || echo "$VERSION_TAG")"
   fi
   if [ "$#" == "1" ] && [ "$1" == "long" ]; then
     pushd "$FASTSURFER_HOME/checkpoints" > /dev/null || return
-    VERSION_INFO=$(printf "%s\ncheckpoints:\n%s" "$VERSION_INFO" "$(md5sum -- *)")
+      VERSION_INFO=$(printf "%s\ncheckpoints:\n%s" "$VERSION_INFO" "$(md5sum -- *)")
```
m-reuter: is this indentation on purpose?

pyproject.toml
```diff
@@ -0,0 +1,65 @@
+[build-system]
+requires = ['setuptools >= 61.0.0']
+build-backend = 'setuptools.build_meta'
+
+[project]
+name = 'fastsurfer'
+version = "2.1.1"
```
m-reuter: Should be 2.2.0-dev.

run_fastsurfer.sh
dkuegler: yes, to show the pushd / popd context

dkuegler: but it was pretty inconsistent :(

dkuegler:
The disadvantage of 2.2.0-dev is that with git hash this looks a bit more ugly: 2.2.0-dev-62df19 vs. 2.2.0dev-62df19
I think that is why without the extra dash is usually the convention in all open source projects that i like, e.g. pytorch with stuff like 1.14a3 or Firefox with 63.1rc1.

run_fastsurfer.sh
```diff
@@ -99,7 +100,8 @@ FLAGS:
                             installed already
   --sid <subjectID>       Subject ID to create directory inside \$SUBJECTS_DIR
   --sd  <subjects_dir>    Output directory \$SUBJECTS_DIR (or pass via env var)
-  --t1  <T1_input>        T1 full head input (not bias corrected). Requires an ABSOLUTE Path!
+  --t1  <T1_input>        T1 full head input (not bias corrected). Requires an
+                            ABSOLUTE Path!
```
LeHenschel, line 104: Why is this on a new line?

LeHenschel: Looks ok to me. Did you test if this works correctly in docker and native install?

dkuegler: Newline because of line length

dkuegler:
@LeHenschel finally got the comparison scripts to process everything properly and as I ran both the currently published and this version locally all segmentations and stats files are identical.

It might be that some results differ based on the PC, since @AhmedFaisal95 's sample reference_data has differences to my results.

PR #362, opened 2023-08-14
Adding FreeSurfer commands (for validation tests)

af-a:
## Description

This PR includes a few additional commands and binaries in the pruned Freesurfer install that are used in the FastSurfer validation pipeline.

The validation pipeline is being modified to run FastSurfer processing and Freesurfer commands through the FastSurfer Docker image. Therefore, the following Freesurfer binaries/scripts and their required files have been included:
* `asegstats2table`
* `aparcstats2table`
* `mris_preproc`
* `mri_glmfit`

This increases the FastSurfer Docker image size by ~60MB.

**Note:** 
* The stats Python scripts necessitate the inclusion of Freesurfer's packaged Python interpreter and some libs to work. These account for nearly all of the size increase (> 57MB).
* The Dockerfile was changed to not delete/overwrite Freesurfer's Python.

## Secondary Dependencies

This section lists the binaries, scripts, and files that are needed by the aforementioned Freesurfer binaries/scripts.

### `asegstats2table` and `aparcstats2table`

Python packages:
* `fsbindings`

### `mris_preproc`

Binaries:
* `fname2stem`
* `fspython`
* `mri_surf2surf`

Scripts (without the following, the script still runs through but with silent failures that lead to incorrect results):
* `isanalyze`
* `isnifti`

### `mri_glmfit`

None.

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -175,7 +175,7 @@ COPY /Docker/python-s /venv/bin/
 
 # Copy fastsurfer over from the build context and add PYTHONPATH
 COPY . /fastsurfer/
-ENV PYTHONPATH=/fastsurfer:$PYTHONPATH \
+ENV PYTHONPATH=/fastsurfer:/opt/freesurfer/python/package:$PYTHONPATH \
     FASTSURFER_HOME=/fastsurfer
 
 # Download all remote network checkpoints already, compile all FastSurfer scripts into bytecode and update
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -114,7 +114,9 @@ SHELL ["/bin/bash", "--login", "-c"]
 
 # install freesurfer and point to new python location
 RUN /install/install_fs_pruned.sh /opt --upx && \
-    rm -R /install
+rm /opt/freesurfer/bin/fspython &&  \
+rm -R /install && \
+ln -s /venv/bin/python3 /opt/freesurfer/bin/fspython
 
 
 # =======================================================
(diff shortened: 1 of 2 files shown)


af-a:
## UPDATE

The Freesurfer Python interpreter and other extra files were removed; only the stats scripts' non-native dependency `fsbindings.legacy` is now included in the Freesurfer Python files. 

For this to work, the `fsbindings` `__init__.py` is edited in `install_fs_pruned.sh` to make the submodule accessible without needing the pre-compiled library (.so file), which would have required Freesurfer's Python version to run.

The following is an example command for running `asegstats2table` using a pre-built FastSurfer Docker image:
```
docker run --gpus=all --rm --name fastsurfer_asegstats \
	-v ${freesurfer_dir}:/freesurfer_dir -v ${subject_list_dir}:/subject_list_dir \
	-v ${subjects_dir}:/subjects_dir -v ${output_dir}:/output_dir 
	-e FS_LICENSE="/freesurfer_dir/.license" -e SUBJECTS_DIR="/subjects_dir" --entrypoint asegstats2table \
	--user $(id -u):$(id -g) fastsurfer:gpu \
	--subjectsfile=/subject_lists_dir/${subject_list_file} --meas=volume \
	--tablefile=/output_dir/aseg.${run_id}.volume.txt --delimiter=space --stats=aseg.stats
```

This modification reduces the additional image size from ~60MB to ~40MB.

-------

EDIT: Removed an incorrect comment that "the `asegstats2table` and `aparcstats2table` commands (shell scripts) can not be used directly, because they explicitly call the Python scripts using Freesurfer's Python interpreter. Instead, the user must run the Python scripts. "

Issue #378, opened 2023-10-05
Incorrect eTIV estimates


dkuegler:
## Description
We have found and confirmed too-slow eTIV estimates, for example as found in <subject_dir>/<subject_id>/stats/aseg.stats.
eTIV estimates are in some cases too small (volumes significantly below 1.5L).

This happens, because the talairach registration "fails" (it does not "fail-safe", but produces unreasonable registration matrices).
Since the eTIV estimation is based on the talairach registration, eTIV values are affected.

Both talairach registration and eTIV computation is performed by FreeSurfer binaries.

Note, that limitations to FreeSurfers eTIV estimation are well-known and documented, but we did not previously (and do not at this moment) specifically know how often this issue arises for FastSurfer.

## Steps to Reproduce
Run FastSurfer (full pipeline) on affected subjects.

## Expected Behavior
reasonable/correct eTIV estimates

## Environment
any.

## QC-Recommendations
Currently, etiv estimates of fastsurfer cannot be trusted.

At a minimum, check all eTIV values. 
Specifically, the Measure BrainSegVol-to-eTIV in the aseg.stats file should be smaller than 0.9

## Solutions
We will:
1. Gather and clarify statistics about the reported eTIV values.
2. Introduce a fix ASAP, likely based on a reasonable initialization of the talairach-registration.

robbwh:
You probably already know about this option, but the flag "-samseg-reg" in recon-all for talairach registration is superior in my experience.

dkuegler:
My initial analysis shows, that etiv estimation error is relatively prevalent. 

A BrainSegVol-to-eTIV < 0.9 is very indicative of a failed talairach registration.

Before this issue is fixed, eTIV estimates of FastSurfer should not be trusted.

dkuegler:
@robbwh Thanks!

We are aware of the option, but this cannot easily be added to FastSurfer at the moment. It does point out an alternative approach to etiv calculation (that we were also aware of :-) ) as well ... through the samseg segmentation.

m-reuter:
A consistent shift in eTIV is possible and no reason for concern (since we register a different image to the talairach average). Did you check the registration visually? We also need to make sure that for this check you use the correct talairach average as FreeSurfer has several. 
Further, how many cases did you find where the registration failed? Does this happen for a specific dataset or resolution?

m-reuter:
[...]

PR #387, opened 2023-11-09
fix the global normalization in bias field correction

dkuegler:
To address incorrect etiv estimates, several changes to the bias field estimation and the pipeline are implemented
- use the white matter mask of the aseg to identify the rescaling slope so images have a predefined white matter intensity.
- add ifh2hdr from Freesurfer to the docker image (used by talairach_avi)
- change the directory the bashrc is written to in the docker image
- general cleanup and formatting of n4_bias_correct.py
- add overloads to image_io.py
- fix python-s reference when it does not exist (the idea is that in the docker, packages from user-space , so a mounted home directory, should never be used)

Addresses #378 .

### DIFF
--- a/recon_surf/N4_bias_correct.py
+++ b/recon_surf/N4_bias_correct.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python3
-
-
-# Copyright 2019 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
+# Copyright 2023 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,14 +15,18 @@
 
 
 # IMPORTS
-import optparse
+import argparse
 import sys
-from typing import Any, Optional
+from pathlib import Path
+from typing import Optional, cast
+import logging
+
 import SimpleITK as sitk
 import numpy as np
 from numpy import typing as npt
 
 import image_io as iio
+
 
 HELPTEXT = """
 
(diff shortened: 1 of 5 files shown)


run_fastsurfer.sh
```diff
@@ -730,7 +734,7 @@ if [ "$run_seg_pipeline" == "1" ]
       then
         # this will always run, since norm_name is set to subject_dir/mri/orig_nu.mgz, if it is not passed/empty
         echo "INFO: Running N4 bias-field correction" | tee -a "$seg_log"
-        cmd="$python ${reconsurfdir}/N4_bias_correct.py --in $conformed_name --out $norm_name --mask $mask_name --threads $threads"
+        cmd="$python ${reconsurfdir}/N4_bias_correct.py --in $conformed_name --out $norm_name --mask $asegdkt_segfile --threads $threads"
```
m-reuter: Why was this changed to asegdkt instead of mask ?

m-reuter: maybe this should be the --aseg flag?

recon_surf/N4_bias_correct.py
```diff
@@ -177,257 +184,388 @@ def N4correctITK(
...
 
-    if not options.skipwm:
-        print("\nnormalize WM to 110")
-        itkcorrected = normalizeWM(itkcorrected, itkmask, centroid=talorig)
+    target_wm = 110.
+
+    if options.skipwm:
+        logger.info("Skipping WM normalization, ignoring talairach and aseg inputs")
+        # normalize to average input intensity
+        m_bf_img = get_image_mean(itk_bfcorr_image)
```
m-reuter: Maybe here we should get the means in the mask (if it is passed) as we do not care much about the intensities outside the brain.

dkuegler:
@m-reuter The call to N4_bias_correct was in fact wrong.
I fixed that now. I tested the script not with the run_fastsurfer.sh script, but directly that script, so it should still be correct, it just means, that my tests/validation for the etiv values and the PV-effects are based on the wrong script...
using --mask was a version that I considered, but then I changed it again because it was potentially confusing and not clear.

So, unfortunate... It is going to delay any new release...

run_fastsurfer.sh
dkuegler: yes, it should :(

recon_surf/N4_bias_correct.py
dkuegler: I think it probably makes little difference, but we can certainly do that...

PR #388, opened 2023-11-09
add scripts for Fastsurfer orchestration on slurm clusters

dkuegler:
Includes the bias field commits

Adds three files to run Fastsurfer on a subject_list file and on a slurm cluster in a resource-optimized way.

Known issues:
On our slurm cluster I experienced some sporadic out-of-memory issues where they should not have happened. These need to be investigated.


m-reuter:
Nice! I wonder if it makes sense to move these into a slurm subdirectory or into the singularity subdir?

PR #389, opened 2023-11-16
Bump versions of python packages

dkuegler:
Bump version of python packages

Also one "forgotten" change of the bias field (in recon surf)


dkuegler:
As an fyi, pip-compile is used to generate the requirements.txt file, which can theoretically also use the pyproject.toml file, but currently does not work.

I went through the @dependabot PRs and looked at the versions. Only pillow is directly mentioned in our reuqirements, all the other packages are derived (i.e. those are versions determined by pip).

Versions from PRs
```
pillow 10.0.1
werkzeug 3.0.1
urllib3 1.26.18
certifi 2023.7.22
grpcio 1.53.0
oauthlib 3.2.1
wheel 0.38.1
```

m-reuter:
yes, did you create a new requirements.txt from the ..in file with updated requirements? Because there also only the pillow version changed?

m-reuter:
also probably the requirements_cpu.txt needs update.

Issue #418, opened 2023-12-20
Is the package yml located in the top level?


pmahon3:
https://github.com/Deep-MI/FastSurfer/blame/a000faa1a5b9fd2355b6d9e8502c64929dd0ae07/INSTALL.md#L96

As per the native install instructions the package yml should be in the top level of the repo but it doesn't seem to be there. The only yaml's i see are under /env https://github.com/Deep-MI/FastSurfer/tree/2118c5c2104874c7bdef1243e4c1de31fd2ffbd7/env

dkuegler:
Hi,

That is indeed outdated with the most recent release. We no longer have multiple environments per hardware setup either. 

The short version is: use `env/FastSurfer.yaml` now (for conda) or the requirements.txt in root (for pip).

I''ll update the docs.

dkuegler:
Maybe I should double-check if the recon-surf ymls file is actually the same as the `MODE=base` file. Then we could consider dropping the reconsurf yml file...

PR #421, opened 2023-12-23
fix native install doc

dkuegler:
Conda environment files have moved and non-cuda files are generated on the fly.

As reported by #418 .

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,12 +93,12 @@ cd FastSurfer
 Create a new environment and install FastSurfer dependencies:
 
 ```bash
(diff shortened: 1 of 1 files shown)


dkuegler:
Maybe I should double-check if the recon-surf yml file is actually the same as the `MODE=base` file. Then we could consider dropping the reconsurf yml file...

dkuegler:
I double-checked. It is not exactly the same, so there are some packages that are installed that might not be needed for recon_surf. In the end, it requires testing... Which I can do after Christmas...

Issue #436, opened 2024-01-23
question about subject list in FastSurfer on multiple subjects (usgae of --subject_list)


caotihutang:
## Question/Support Request

I try lots of formats for subjects list file, but it still reported an error. the subjects list is below.
subject1=--t1 /data/subjectx/001.nii.gz\n
subject2=--t1 /data/subjectx/002.nii.gz\n
subject3=--t1 /data/subjectx/003.nii.gz\n
subject4=--t1 /data/subjectx/004.nii.gz\n
subject5=--t1 /data/subjectx/005.nii.gz\n
subject6=--t1 /data/subjectx/006.nii.gz\n

and I also try to exclude the "--t1" like this
subject1=/data/subjectx/001.nii.gz\n
subject2=/data/subjectx/002.nii.gz\n
subject3=/data/subjectx/003.nii.gz\n
subject4=/data/subjectx/004.nii.gz\n
subject5=/data/subjectx/005.nii.gz\n
subject6=/data/subjectx/006.nii.gz\n

## Screenshots

![image](https://github.com/Deep-MI/FastSurfer/assets/48438844/f6e5ca40-5483-4013-a957-bfdcf2211340)

## Environment
 - FastSurfer Version: cpu-lastest
 - FreeSurfer Version: ... 
 - OS: Windows 10
 - CPU: ...
 - GPU: ...

### Execution

Run Command:

dkuegler:
As you can see in "Reading subjects from stdin", the script does not read the subjects at all. Its not the format, its the subject list.
if you look at `brun_fastsurfer.sh --help`, you can see that your `--subjects_list` is incorrect.

```
brun_fastsurfer.sh --subject_list <file> [other options]
```

I realize that we have a typo in the docs there, I'll fix that.... I had some inconsistent usage of --subject_list and --subjects_list.... this one must have slipped through.

Thanks for pointing this out.

caotihutang:
thanks

PR #437, opened 2024-01-23
Fix subjects_list typo in README.md

dkuegler:
should be --subject_list.


dkuegler:
Addresses issue raised in #436 .

PR #445, opened 2024-01-31
Add Docker attestation and provenance

dkuegler:
Dockerfile
- change user to nonroot
- add the BUILDKIT_SBOM_SCAN_CONTEXT buildarg for proper SBOM creation Docker/build.py
- add provenance
- add sbom attestation
- some formattíng

This is so the FastSurfer docker image satisfies docker scout criteria.

At this point, it might be necessary to check whether the user inside the docker container is manually entered, or otherwise weird things happen (similar to our --allow-root flag).
I will look into that.

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -123,8 +123,10 @@ FROM build_base AS build_freesurfer
 COPY ./Docker/install_fs_pruned.sh /install/
 SHELL ["/bin/bash", "--login", "-c"]
 
+ARG FREESURFER_URL=default
+
 # install freesurfer and point to new python location
-RUN /install/install_fs_pruned.sh /opt --upx && \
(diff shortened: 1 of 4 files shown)


dkuegler:
I hope this finally works.... unfortunately, the attestation changes mean that people that want to do attestation, need to change a couple of things, such as:
- change to the containerd image store
- set up a docker-container buildx builder

but the basic build should still work without that, just attestation are a bit weird :(.

m-reuter:
please add some documentation to readme, including attestation requirements.

dkuegler:
Documentation added to Docker/README.md and an additional option that does not require containerd, but a local registry.

Unfortunately, the sbom is incomplete as most the version of Freesurfer as well as Fastsurfer are not included.

PR #457, opened 2024-02-14
Swapping out use of conda in building the dockerfile for the much faster mamba

jrussell9000:
Replacing use of conda in the Dockerfile (and conda_pack.sh) with mamba. Dramatically reduces build time.

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -23,14 +23,14 @@
 #   Image to use to install freesurfer binaries from, the freesurfer binaries
 #   should be located in /opt/freesurfer in the image.
 #   - default: build_freesurfer
-# - CONDA_BUILD_IAMGE:
+# - CONDA_BUILD_IMAGE:
 #   Image to use to install the python environment from, the python environment
 #   should be in /venv/ in the image.
 #   - default: build_cuda
-# - CONDA_FILE:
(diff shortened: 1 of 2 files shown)


dkuegler:
This looks like a nice update.

We are usually building with build kit and just caching the conda image, so it's a one-time thing to build, but mamba will speed up this part of  the build.... For us, the freesurfer build usually takes longer...

There is a small change that we would still need 
We need a fixed version of mamba (not latest, also currently it always has to choose python 3.10). Probably MAMBA_VERSION instead of MAMBA_FILE as that does not seem to change for Miniforge.

After that, we'll have to test the resulting image...

m-reuter:
I fixed the version to conda 23.11 which is currently latest and therefore this bumps the version from 23.9 which we used before. This supports python 3.8 to 3.11 but not sure if we need to change anything else to force it to use version 3.10 ?

dkuegler:
I am not really sure this makes any sense. Mamba is an alternative to conda, which is open source and usually quite a bit faster. I'll see if I can fix this :)

dkuegler:
Also python 3.10 is specified through the fastsurfer environment file in `env/fastsurfer.yml`, so that should be fine.

I'll build the image using this branch tomorrow and then we can see, if it works as expected or not.

PR #465, opened 2024-02-27
 Fix output type of N4 bias field correction and add uchar flag

### LINKED ISSUE #461: N4 bias field correction without rescaling also converts to UCHAR
## Description
In recon_surf/N4_bias_correct.py if it is called to output the --outvol, but the input is not UCHAR (i.e. the mean is not around 150 or so) the script renormalizes the input to the old mean (of the input image). After that, it is converted to UCHAR by clamping, so all intensity information >255 is lost. This is not an expected behavior.

## Steps to Reproduce
Run N4_bias_correct on an image that is not UCHAR / not conformed.

## Expected Behavior
Retains data type, or additional flag for conforming intensities.

dkuegler:
Adds a flag flag and logic to keep/cast dtype of outvol to uchar (which also robustly rescales the intensities).

Fixes #461

### DIFF
--- a/recon_surf/N4_bias_correct.py
+++ b/recon_surf/N4_bias_correct.py
@@ -84,12 +84,13 @@ Original Author: Martin Reuter
 Date: Mar-18-2022
 
 Modified: David Kügler
-Date: Oct-25-2023
+Date: Feb-27-2024
 """
 
 h_verbosity = "Logging verbosity: 0 (none), 1 (normal), 2 (debug)"
 h_invol = "path to input.nii.gz"
 h_outvol = "path to corrected.nii.gz"
+h_uchar = "sets the output dtype to uchar (only applies to outvol, rescalevol is uchar by default.)"
 h_rescaled = "path to rescaled.nii.gz"
 h_mask = "optional: path to mask.nii.gz"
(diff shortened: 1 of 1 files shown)


dkuegler:
@santiestrada32 can you check if this solves the situation with bias field correction in #459 ? The image looks fine to me.

`N4_bias_correct-py --in <file> --out <outfile> --threads 12 --uchar`

m-reuter:
Also make sure that you adjust the places where this is called accordingly (in run_fastsurfer and in recon-surf).

dkuegler:
Those places do not need to be adjusted because they are not affected. There, we do not use the outvol variable (bias field before wm normalization) which is the change here. The issue popped up, because for T2, WM normalization does not happen.

m-reuter:
just thought that adding uchar flag there makes sense, given that we also want UCHAR output in those places? Even if it is the default but it will be easier to see?

dkuegler:
I did not want to change the behavior where it was not necessary. I guess we have just changed that script in 2.2 so we can change the arguments again. But I am not really certain if there is ever a situation, where you would want the WM normalized image to not be UCHAR.

santiestrada32:
I check the new changes in my dataset, and everything looks okay

Issue #464, opened 2024-02-25
python3.10 command not found: error running ./run_fastsurfer.sh


suhanasaigoli05:
running: 
./run_fastsurfer.sh --t1 $datadir/bert/mri/T1.mgz \
                    --sid bert --sd $fastsurferdir \
                    --parallel --threads 4 --3T
error:
<img width="1066" alt="image" src="https://github.com/Deep-MI/FastSurfer/assets/63617775/def5dc1d-e13f-4134-b120-f2d59ee51770">

## Environment
 - OS: MacOS
 - Native install

python packages:
<img width="234" alt="image" src="https://github.com/Deep-MI/FastSurfer/assets/63617775/38c6cee0-35c0-450b-b2ce-8dc8fa52e797">
<img width="256" alt="image" src="https://github.com/Deep-MI/FastSurfer/assets/63617775/e7fe54c5-3bfb-4156-ac06-dd6f9a94156d">
<img width="229" alt="image" src="https://github.com/Deep-MI/FastSurfer/assets/63617775/d30fd8ed-b873-4deb-9eac-8d05b0d2dafe">

Run Command: 

export FREESURFER_HOME=//Applications/freesurfer/7.4.1
source $FREESURFER_HOME/SetUpFreeSurfer.sh
datadir=/Applications/freesurfer/7.4.1/subjects
fastsurferdir=/Users/username/Applications/FastSurfer
./run_fastsurfer.sh --t1 $datadir/bert/mri/T1.mgz \
                    --sid bert --sd $fastsurferdir \
                    --parallel --threads 4 --3T

dkuegler:
Hi,

It seems you are running python 3.11.4. Fastsurfer 2.2 only officially supports python 3.10., but you should still be able to run Fastsurfer with 3.11 by adding the flag `--py python3.11` though.

Although our recommendation would be to set up an environment with python 3.10...

m-reuter:
Could also be an old bash version or the "tee" command ?

m-reuter:
closing this for now, feel free to re-open if this does not solve the issue.

PR #466, opened 2024-02-28
Split talairach registration into recon_surf/talairach-reg.sh and python not found error

dkuegler:
run_fastsurfer.sh
- check for existence of python with error message
- move recon-surf flags into surf_flags array recon_surf/recon-surf.sh
- remove talariach registration from script 
Docker/build.py
- fix build error without attestation

### DIFF
--- a/run_fastsurfer.sh
+++ b/run_fastsurfer.sh
@@ -60,14 +60,7 @@ run_asegdkt_module="1"
 run_cereb_module="1"
 threads="1"
 # python3.10 -s excludes user-directory package inclusion, but passing "python3.10 -s" is not possible
-# python-s is a miniscript to add this flag, but this only works if python-s is defined
-#if [[ -n "$(which python-s)" ]]; then
-  python="python3.10 -s"
-#elif [[ -f "/fastsurfer/python-s" ]]; then
-#  python="/fastsurfer/python-s"
-#else
-#  python="python3.10"
-#fi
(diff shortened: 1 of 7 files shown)


dkuegler:
This version is tested for three settings, `--seg_only`, `--seg_only --tal_reg` as well as `--tal_reg`. Seems to be working for all :).

dkuegler:
This PR also provides better error managing if the python version does not match as reported in #464 .

run_fastsurfer.sh
```diff
@@ -60,14 +60,7 @@ run_asegdkt_module="1"
 run_cereb_module="1"
 threads="1"
 # python3.10 -s excludes user-directory package inclusion, but passing "python3.10 -s" is not possible
```
m-reuter: here it still says that python.. -s is not possible, but below you define it like that?

dkuegler: Yeah, I fixed that in this commit

m-reuter: No, you did not fix this comment , Line 62, which is still in here.,

dkuegler: What I meant was
This comment is outdated. It used to be true, but this PR fixes the underlying issue

dkuegler:
```suggestion
# python3.10 -s excludes user-directory package inclusion
```

PR #467, opened 2024-02-29
plane defaults

agirodi:
This PR adds:
* Backup URL feature, where the checkpoint download can have multiple URLs, in case the first one is unreachable or similar
* New yaml files ([CerebNet/config/checkpoint_paths.yaml](CerebNet/config/checkpoint_paths.yaml) and [FastSurferCNN/config/checkpoint_paths.yaml](FastSurferCNN/config/checkpoint_paths.yaml)) for default values for the configuration path, the checkpoint path as well as the URLs

### DIFF
--- a/CerebNet/run_prediction.py
+++ b/CerebNet/run_prediction.py
@@ -22,12 +22,13 @@ from pathlib import Path
 from FastSurferCNN.utils import logging, parser_defaults
 from CerebNet.utils.load_config import get_config
 from CerebNet.inference import Inference
-from FastSurferCNN.utils.checkpoint import get_checkpoints
+from FastSurferCNN.utils.checkpoint import get_checkpoints, get_plane_default
 from FastSurferCNN.utils.common import assert_no_root, SubjectList
+from FastSurferCNN.utils.parser_defaults import FASTSURFER_ROOT
 
 logger = logging.get_logger(__name__)
 DEFAULT_CEREBELLUM_STATSFILE = Path("stats/cerebellum.CerebNet.stats")
-
+CEREBNET_CHECKPOINT_PATHS_FILE = FASTSURFER_ROOT / "CerebNet/config/checkpoint_paths.yaml"
 
 def setup_options():
     """
@@ -80,20 +81,14 @@ def setup_options():
         ["device", "viewagg_device", "threads", "batch_size", "async_io", "allow_root"],
     )
 
-    from CerebNet.utils.checkpoint import CEREBNET_COR, CEREBNET_AXI, CEREBNET_SAG
 
     parser_defaults.add_plane_flags(
         advanced,
         "checkpoint",
-        {"coronal": CEREBNET_COR, "axial": CEREBNET_AXI, "sagittal": CEREBNET_SAG},
+        {"coronal": "default", "axial": "default", "sagittal": "default"},
+        CEREBNET_CHECKPOINT_PATHS_FILE
     )
 
-    parser.add_argument(
-        "--cfg",
-        dest="cfg_file",
-        help="Path to the config file",
-        type=Path,
-    )
     parser.add_argument(
         "opts",
         help="See CerebNet/config/cerebnet.py for additional options",
@@ -138,7 +133,6 @@ def main(args: argparse.Namespace) -> int | str:
 
     # Set up logging
     from FastSurferCNN.utils.logging import setup_logging
-    from CerebNet.utils.checkpoint import URL as CEREBNET_URL
 
     setup_logging(getattr(args, "log_name"))
 
@@ -157,7 +151,10 @@ def main(args: argparse.Namespace) -> int | str:
         subjects_kwargs["norm_name"] = "norm_name"
 
     logger.info("Checking or downloading default checkpoints ...")
-    get_checkpoints(args.ckpt_ax, args.ckpt_cor, args.ckpt_sag, url=CEREBNET_URL)
+    
+    urls = get_plane_default("URL", filename=CEREBNET_CHECKPOINT_PATHS_FILE)
+
+    get_checkpoints(args.ckpt_ax, args.ckpt_cor, args.ckpt_sag, urls=urls)
 
     # Check input and output options and get all subjects of interest
     subjects = SubjectList(
@@ -179,4 +176,6 @@ def main(args: argparse.Namespace) -> int | str:
 
 if __name__ == "__main__":
     parser = setup_options()
-    sys.exit(main(parser.parse_args()))
+    args = parser.parse_args()
+
+    sys.exit(main(args))
--- a/FastSurferCNN/README.md
+++ b/FastSurferCNN/README.md
@@ -20,14 +20,14 @@ The *FastSurferCNN* directory contains all the source code and modules needed to
 * `--strip`: strip suffix from path definition of input file to yield correct subject name. (Optional, if full path is defined for `--t1`)
 * `--lut`: FreeSurfer-style Color Lookup Table with labels to use in final prediction. Default: ./config/FastSurfer_ColorLUT.tsv
 * `--seg`: Name of intermediate DL-based segmentation file (similar to aparc+aseg).
-* `--cfg_cor`: Path to the coronal config file
-* `--cfg_sag`: Path to the axial config file
-* `--cfg_ax`: Path to the sagittal config file
 
-#### Checkpoints
+#### Checkpoints and configs
 * `--ckpt_sag`: path to sagittal network checkpoint
 * `--ckpt_cor`: path to coronal network checkpoint
 * `--ckpt_ax`: path to axial network checkpoint
+* `--cfg_cor`: Path to the coronal config file
+* `--cfg_sag`: Path to the axial config file
+* `--cfg_ax`: Path to the sagittal config file
 
 #### Optional commands
 * `--clean`: clean up segmentation after running it (optional)
--- a/FastSurferCNN/run_prediction.py
+++ b/FastSurferCNN/run_prediction.py
@@ -18,37 +18,41 @@ import copy
 import sys
 from concurrent.futures import Executor, ThreadPoolExecutor, Future
 from pathlib import Path
-from typing import Any, Iterator, Literal, Optional, Sequence
+from typing import Dict, Any, Iterator, Literal, Optional, Sequence
 
 import nibabel as nib
 import numpy as np
 import torch
+import nibabel as nib
 import yacs.config
 
 import FastSurferCNN.reduce_to_aseg as rta
 from FastSurferCNN.data_loader import conform as conf
 from FastSurferCNN.data_loader import data_utils as du
 from FastSurferCNN.inference import Inference
-from FastSurferCNN.quick_qc import check_volume
 from FastSurferCNN.utils import logging, parser_defaults
-from FastSurferCNN.utils.checkpoint import VINN_AXI, VINN_COR, VINN_SAG, get_checkpoints
+from FastSurferCNN.utils.checkpoint import get_checkpoints, get_plane_default
+from FastSurferCNN.utils.load_config import load_config
 from FastSurferCNN.utils.common import (
     SerialExecutor,
     SubjectDirectory,
     SubjectList,
     assert_no_root,
     find_device,
+    assert_no_root,
     handle_cuda_memory_exception,
+    SubjectList,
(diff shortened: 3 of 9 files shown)


plane_default.yaml
dkuegler, line 1: This file should probably be called `checkpoint_paths.yaml` or similar. Also, it should be located in `FastSurferCNN/config`.

plane_default.yaml
```diff
@@ -0,0 +1,21 @@
+DOWNLOADS:
+  URL: "https://b2share.fz-juelich.de/api/files/a423a576-220d-47b0-9e0c-b5b32d45fc59"
+  URL_BACKUP: "https://zenodo.org/records/10390573/files"
```
dkuegler:
```suggestion
  URL:
  - "https://b2share.fz-juelich.de/api/files/a423a576-220d-47b0-9e0c-b5b32d45fc59"
  - "https://zenodo.org/records/10390573/files"
```

This probably causes some other changes as it is a list now

plane_default.yaml
```diff
@@ -0,0 +1,21 @@
...
+  CKPT:
+    AXIAL: "FastSurferCNN/checkpoints/aparc_vinn_axial_v2.0.0.pkl"
+    CORONAL: "FastSurferCNN/checkpoints/aparc_vinn_coronal_v2.0.0.pkl"
+    SAGITTAL: "FastSurferCNN/checkpoints/aparc_vinn_sagittal_v2.0.0.pkl"
+  CFG:
+    AXIAL: "FastSurferCNN/config/FastSurferVINN_axial.yaml"
+    CORONAL: "FastSurferCNN/config/FastSurferVINN_coronal.yaml"
+    SAGITTAL: "FastSurferCNN/config/FastSurferVINN_sagittal.yaml"
+
+CEREBNET:
```
dkuegler: Maybe the CerebNet paths can also  be in a different file, if you want. But the URL for CerebNet is also different.

FastSurferCNN/README.md
```diff
@@ -20,14 +20,15 @@ The *FastSurferCNN* directory contains all the source code and modules needed to
...
 
-#### Checkpoints
+#### Checkpoints and configs
 * `--ckpt_sag`: path to sagittal network checkpoint
 * `--ckpt_cor`: path to coronal network checkpoint
 * `--ckpt_ax`: path to axial network checkpoint
+* `--cfg_cor`: Path to the coronal config file
+* `--cfg_sag`: Path to the axial config file
+* `--cfg_ax`: Path to the sagittal config file
+* `--plane_yaml` Path to the yaml file, containing url, ckpt and cfg information
```
dkuegler: I am not sure about making this a parameter .... these are default values.

CerebNet/run_prediction.py
```diff
@@ -17,18 +17,45 @@
...
 from CerebNet.inference import Inference
-from FastSurferCNN.utils.checkpoint import get_checkpoints
+from FastSurferCNN.utils.checkpoint import get_checkpoints, get_plane_dict, PLANE_DEFAULTS
 from FastSurferCNN.utils.common import assert_no_root, SubjectList
 
 logger = logging.get_logger(__name__)
 DEFAULT_CEREBELLUM_STATSFILE = "stats/cerebellum.CerebNet.stats"
 
 
+def merge_dict_in_args(args: argparse.Namespace, dictionary: Dict, defaults: Dict) -> argparse.Namespace:
```
dkuegler:
```suggestion
def merge_dict_in_args(args: argparse.Namespace, dictionary: dict, defaults: dict) -> argparse.Namespace:
```

CerebNet/run_prediction.py
```diff
@@ -17,18 +17,45 @@
 # IMPORTS
 import sys
 import argparse
+from typing import Dict
```
dkuegler:
```suggestion
```

FastSurferCNN/utils/parser_defaults.py
dkuegler, line 1: I would prefer it, if the default values are loaded (say by putting a load function as the default value for files in add_plane_flags, there the parameter to add_plane flags is the file name from which to get the values).

CerebNet/run_prediction.py
```diff
@@ -75,20 +102,28 @@ def setup_options():
...
 
     parser.add_argument(
         "--cfg",
         dest="cfg_file",
         help="Path to the config file",
         type=str,
+        default=PLANE_DEFAULTS["CEREBNET"]["CFG"]["UNIVERSAL"],
+    )
+
+    parser.add_argument(
```
dkuegler: I dont think this config file should be a parameter to the script.

CerebNet/run_prediction.py
```diff
@@ -17,18 +17,45 @@
...
+                wrong_files.append(f"cfg_file")
+
+
+    if len(wrong_files) > 0:
+        raise ValueError("The following files were given twice, once in the command line and once in the yaml file, but do not match:"
+            + ", ".join(wrong_files))
+
+    return args
+
 def setup_options():
```
dkuegler, line 33: For the loading to work properly, this function needs to return a custom Parser, which in turn implements 'parse_known_args'.

```python
class CustomParser(argparse.ArgumentParser):
    def parse_known_args(self, args=None, namespace=None):
        args, argv = super().parse_known_args(args, namespace)
        # ... here update arguments that are needed to be updated from the config file
        return args, argv
```

CerebNet/run_prediction.py
```diff
@@ -75,20 +102,28 @@ def setup_options():
         ["device", "viewagg_device", "threads", "batch_size", "async_io", "allow_root"],
     )
 
-    from CerebNet.utils.checkpoint import CEREBNET_COR, CEREBNET_AXI, CEREBNET_SAG
 
     parser_defaults.add_plane_flags(
         advanced,
         "checkpoint",
-        {"coronal": CEREBNET_COR, "axial": CEREBNET_AXI, "sagittal": CEREBNET_SAG},
+        {"coronal": PLANE_DEFAULTS["CEREBNET"]["CKPT"]["CORONAL"], "axial": PLANE_DEFAULTS["CEREBNET"]["CKPT"]["AXIAL"], "sagittal": PLANE_DEFAULTS["CEREBNET"]["CKPT"]["SAGITTAL"]},
```
dkuegler: maybe just "__default__" or a a function that updates on-demand here?

dkuegler (commented):
I would like this to be done a bit differently.

FastSurferCNN/README.md
agirodi: The idea was, that it could be easier to pass a file with other ckps and cfgs, as an alternative to the somewhat long flags. But this might be a niche case and not neccesarily adds a lot of benefit. I could revert it back if you want.

dkuegler: But the information in the yaml file (the file we pass here) is all static information. There is no reason to change this for users.
This change primarily for us to dis-aggregate code and content/configuration/data.

CerebNet/run_prediction.py
agirodi: Can you elaborate why the loading would not otherwise work properly unless we use a custom parser class instead of the usual one with added arguments? Also this sounds like a different feature that goes beyond this change of the default ckpt value. I could work on it, but I would not include it in this PR

dkuegler: basically, you want to modify the behavior of how arguments are parsed and replaced, you need that do happen somewhere. This can be done by Actions (from argparse... which can be custom Actions), but here, you want it to be done after all parameters are parsed. So after all parameters are parsed, you want to replace some values if they are still the default -- including a check if the service can be reached and using the backup, if they main server is unreachable.

FastSurferCNN/run_prediction.py
```diff
@@ -18,37 +18,41 @@
 import sys
 from concurrent.futures import Executor, ThreadPoolExecutor, Future
 from pathlib import Path
-from typing import Any, Iterator, Literal, Optional, Sequence
+from typing import Dict, Any, Iterator, Literal, Optional, Sequence
 
 import nibabel as nib
 import numpy as np
 import torch
+import nibabel as nib
```
dkuegler:
```suggestion
```

FastSurferCNN/run_prediction.py
```diff
@@ -18,37 +18,41 @@
 import sys
 from concurrent.futures import Executor, ThreadPoolExecutor, Future
 from pathlib import Path
-from typing import Any, Iterator, Literal, Optional, Sequence
+from typing import Dict, Any, Iterator, Literal, Optional, Sequence
```
dkuegler:
```suggestion
from typing import Any, Iterator, Literal, Optional, Sequence
```

FastSurferCNN/utils/checkpoint.py
```diff
@@ -12,28 +12,81 @@
...
+    """
+    Load the plane dictionary from the yaml file.
+
+    Parameters
+    ----------
+    filename : Path
+        path to the yaml file. Either absolute or relative to the FastSurfer root directory.
+    Returns
+    -------
+
```
dkuegler:
```suggestion
```

FastSurferCNN/utils/checkpoint.py
```diff
@@ -12,28 +12,81 @@
...
+def get_plane_dict(filename : Path = YAML_DEFAULT) -> dict[str, Union[str, list[str]]]:
+    """
+    Load the plane dictionary from the yaml file.
+
+    Parameters
+    ----------
+    filename : Path
+        path to the yaml file. Either absolute or relative to the FastSurfer root directory.
+    Returns
```
dkuegler, line 45:
```suggestion

    Returns
```

FastSurferCNN/utils/checkpoint.py
```diff
@@ -12,28 +12,81 @@
...
+    Get the default value for a specific plane.
+
+    Parameters
+    ----------
+    type : str
+        Type of value. Can be URL, CKPT, CFG
+    plane : str
+        Plane to get the default value for. Can be "axial", "coronal" or "sagittal".
+    filename : str | Path
+        path to the yaml file. Either absolute or relative to the FastSurfer root directory.
```
dkuegler:
```suggestion
        The path to the yaml file. Either absolute or relative to the FastSurfer root directory.
```

Should start with a capital letter (also no repeating of words).

FastSurferCNN/utils/checkpoint.py
```diff
@@ -12,28 +12,81 @@
...
+    return dictionary
+
+def get_plane_default(type : str, plane : Optional[str] = None, filename : str | Path = YAML_DEFAULT) -> Union[str, list[str]]:
+    """
+    Get the default value for a specific plane.
+
+    Parameters
+    ----------
+    type : str
+        Type of value. Can be URL, CKPT, CFG
```
dkuegler:
```suggestion
def get_plane_default(type : Literal["URL", "CKPT", "CFG"], plane : Optional[str] = None, filename : str | Path = YAML_DEFAULT) -> str | list[str]:
    """
    Get the default value for a specific plane.

    Parameters
    ----------
    type : "URL", "CKPT", "CFG"
        Type of value.
```

FastSurferCNN/utils/checkpoint.py
```diff
@@ -12,28 +12,81 @@
...
+    plane : str
+        Plane to get the default value for. Can be "axial", "coronal" or "sagittal".
+    filename : str | Path
+        path to the yaml file. Either absolute or relative to the FastSurfer root directory.
+
+    Returns
+    -------
+    Union[str, list[str]]
```
dkuegler:
```suggestion
    str, list[str]
```

dkuegler (changes requested):
Please check if the changed I made break anything. Otherwise can be merged.

dkuegler (approved):

dkuegler:
I will squash minor commits in a sec.

FastSurferCNN/utils/parser_defaults.py
```diff
@@ -339,6 +341,7 @@ def add_plane_flags(
     parser: argparse.ArgumentParser,
     type: Literal["checkpoint", "config"],
     files: Mapping[str, Path],
+    default_path: str,
 ) -> argparse.ArgumentParser:
     """
```
m-reuter, line 346:
```suggestion
    default_path: str | Path,
) -> argparse.ArgumentParser:
    """
```

Issue #471, opened 2024-03-05
Issue creating symbolic links when running  fastsurfer-gpu.sif


f4bry:
## Question/Support Request

Dear experts, running `fastsurfer-gpu.sif` I am getting the following error:

```
================== Creating orig and rawavg from input =========================

python-s /fastsurfer/recon_surf/../FastSurferCNN/data_loader/conform.py -i /output/sub-01/mri/orig.mgz --check_only --vox_size min --verbose
Reading input: /output/sub-01/mri/orig.mgz ...
A min-conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        True
 - Voxel Size 1.0x1.0x1.0:        True
 - Orientation LIA:               True
 - Dtype uint8:                   True
Input /output/sub-01/mri/orig.mgz is already conformed! Exiting.

@#@FSTIME  2024:03:06:00:05:46 python-s N 7 e 0.25 S 0.02 U 0.22 P 98% M 41140 F 0 R 6798 W 0 c 5 w 46 I 282 O 0 L 1.56 1.83 1.40
@#@FSLOADPOST 2024:03:06:00:05:47 python-s N 7 1.56 1.83 1.40
python-s /fastsurfer/recon_surf/../FastSurferCNN/data_loader/conform.py -i /output/sub-01/mri/aparc.DKTatlas+aseg.deep.mgz --check_only --vox_size 1.0 --dtype any --verbose
Reading input: /output/sub-01/mri/aparc.DKTatlas+aseg.deep.mgz ...
A conformed image must satisfy the following criteria:
 - Dimensions 256x256x256:        True
 - Voxel Size 1.0x1.0x1.0:        True
 - Orientation LIA:               True
Input /output/sub-01/mri/aparc.DKTatlas+aseg.deep.mgz is already conformed! Exiting.

@#@FSTIME  2024:03:06:00:05:47 python-s N 9 e 0.21 S 0.02 U 0.18 P 98% M 41148 F 0 R 6800 W 0 c 5 w 45 I 210 O 0 L 1.56 1.83 1.40
@#@FSLOADPOST 2024:03:06:00:05:47 python-s N 9 1.56 1.83 1.40
The voxel size 1.0 is not less than 0.999, so we are proceeding with standard options.
mri_convert /output/sub-01/mri/orig.mgz /output/sub-01/mri/orig.mgz
mri_convert /output/sub-01/mri/orig.mgz /output/sub-01/mri/orig.mgz
[...]

PR #473, opened 2024-03-06
Replace softlinks with softlink_or_copy if softlink failed

dkuegler:
Some file systems do not support softlinks, if creating the soft link fails, copy the file instead.

See also #471.

I have no real idea on how to simulate a file system that does not support soft links, i.e. something that makes ln -sf fail, maybe FAT file systems?

### DIFF
--- a/recon_surf/functions.sh
+++ b/recon_surf/functions.sh
@@ -20,11 +20,11 @@ export timecmd
 
 function RunIt()
 {
-# parameters
-# $1 : cmd  (command to run)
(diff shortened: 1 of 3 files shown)


m-reuter:
looks good, have you tried it on a normal system (with soft link) and also the command file version?

dkuegler:
Yeah, the run did not finish earlier, but I can confirm it did finish now.

dkuegler:
Can be merged.

PR #474, opened 2024-03-06
Add brainvol stats to the segstats command

dkuegler:
This PR adds all the brainvol stats measures to segstats.py. This specifically includes eTIV and summary volumes.

It also adds `mri_segstats.py` and `mri_brainvol_stats.py`, which mimic freesurfer binaries, based on this code. They are much faster and use similar command line APIs (as compared to segstats.py, which has a different API).


run_fastsurfer.sh
```diff
@@ -267,246 +273,112 @@ do
...
+  --threads) threads="$1" ; shift ;;
+  --py) python="$1" ; shift ;;
+  -h|--help)
     usage
     exit
     ;;
-    --version)
-    if [[ "$#" -lt 2 ]]; then
+  --version)
+    if [[ "$#" -lt 2 ]] || [[ "$1" =~ ^-- ]]; then
```
dkuegler:
```suggestion
    if [[ "$#" -lt 1 ]] || [[ "$1" =~ ^-- ]]; then
```

CerebNet/inference.py
```diff
@@ -291,6 +291,7 @@ def _get_ids_startswith(_label_map: Dict[int, str], prefix: str) -> List[int]:
         table = pv_calc(
             seg_data,
             norm_data,
+            norm_data,
```
m-reuter, line 294: why do we have norm_date duplicated here?

dkuegler: These are the values for `norm_data` and `pv_guide`. One is used to calculate the intensity values the other to determine the pv corrected volume.

Mri_segstats uses -pv and -norm as equivalent flags.

dkuegler: ... And in CerebNet (here) this is actually the same data (usually also in FreeSurfer, but this is not required)

PR #475, opened 2024-03-07
Sphinx workflow

engrosamaali91:
In order to setup a GitHub Workflow we require gh-pages in fastsurfer repository where html pages will be deployed.


CerebNet/data_loader/augmentation.py
```diff
@@ -333,14 +333,14 @@ def sample_intensity_stats_from_image(
 
     Parameters
     ----------
-    image : array_like
+    image : np.ndarray
```
dkuegler: that is not the same, can you check whether this can be fixed?

CerebNet/utils/metrics.py
```diff
@@ -194,10 +194,10 @@ def hd(result, reference, voxelspacing=None, connectivity=1):
 
     Parameters
     ----------
-    result : array_like
+    result : np.ndarray
```
dkuegler: array_like is not the same as ndarray

.github/workflows/doc.yml
```diff
@@ -0,0 +1,63 @@
...
+        shell: bash
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: ./dev
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
```
dkuegler: I think this should be python 3.10
```suggestion
          python-version: 3.10
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,63 @@
+name: doc
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
+  cancel-in-progress: true
+on:
+  pull_request:
+  push:
+    branches: [dev]
```
dkuegler, line 8: on release, there should probably also be a build of the doc

- dev should always replace the old dev, but release are only ever built once (of the stable branch then).

doc/conf.py
```diff
@@ -7,24 +7,22 @@
...
-sys.path.append(os.path.dirname(__file__) + '/..') 
-sys.path.append(os.path.dirname(__file__) + '/../recon_surf') 
+sys.path.append(os.path.dirname(__file__) + "/..")
+sys.path.append(os.path.dirname(__file__) + "/../recon_surf")
 
-project = 'FastSurfer'
-copyright = '2023, Martin Reuter'
-author = 'Martin Reuter'
+project = "FastSurfer"
+copyright = "2023, Martin Reuter"
```
dkuegler: I don't know how specifically this should be changed, maybe @m-reuter can say, but martin does not have the copyright on "all" of FastSurfer. I would say it makes more sense to say "FastSurfer Team" here or something.

Similarly, there is not just Martin as the author, but a couple of people.

doc/overview/CODE_OF_CONDUCT.md
dkuegler, line 1: It seems these files are added, inmost cases, the correct operation is to move these files. They are also still in the root directory of the project.

doc/overview/Docker/README.md
dkuegler, line 1: This file can stay in folder `Docker` for now.

doc/overview/EDITING.md
dkuegler, line 1: should be moved, not copied

doc/overview/INSTALL.md
dkuegler, line 1: move not copy

doc/overview/README.md
dkuegler, line 1: It looks like this file was copied, but it should not.
1. some of the information in this file should stay in the README.md in the root folder.
2. This is **the** overview file. The other files are in a "overview" folder, which is fine I guess, but they could also be in the the folder `doc`.

So the file that remains in the root folder, should probably have some general information about FastSurfer, but then it can also have several links to files to get more information. I suspect I will have to edit some of this into the `<root>/README.md` file... I guess I have to look into this.

Maybe you can just move all the Examples into its own file, maybe even make dedicated files for each example?
All the information about the flags should be in one file. Another file should be the output files documentation. Then hardware requirements, maybe also modules.

doc/overview/Singularity/README.md
dkuegler, line 1: stays in Singularity

doc/overview/SECURITY.md
dkuegler, line 1: move not copy

dkuegler (changes requested):
Importantly, move not copy.

.github/workflows/doc.yml
m-reuter: currently (in our other repositories) we build on PRs and pushes to our development branch. That way the documentation always reflects the current dev version. Since doc strings are for developers mainly this makes most sense. In order to store (and not replace) webpages for releases, we would need to setup a more complex procedure, where we create a copy of the webpage from dev into a stable directory. Currently this has to be done manually.

doc/overview/README.md
engrosamaali91: The main README.md file is now split into three files
- FASTSURFER_EXAMPLES.md (Contains all six examples)
- FASTSURFER_FLAGS.md (Flags)
- FASTSURFER_OUTFUT_FILES.md

doc/conf.py
m-reuter: Yes, you can say "2024, FastSurfer Developers" . Also I think we do not need an author field here, so that can be dropped.

doc/conf.py
```diff
@@ -21,11 +21,8 @@
 sys.path.append(os.path.dirname(__file__) + "/../recon_surf")
 
 project = "FastSurfer"
-copyright = "2023, Martin Reuter"
-author = "Martin Reuter"
+author = "FastSurfer Developers"
 copyright = f"{date.today().year}, {author}"
```
dkuegler:
```suggestion
copyright = f"2020, {author}"
```

dkuegler (approved):

dkuegler:
1. rebase on upstream
2. Docker and Singularity Files: 
    1. Remove the files in doc/overview
    2. Update the Links in other Markdown files
3. Examples etc.
   1. rename FASTSURFER_EXAMPLES.md to EXAMPLES.md (and other FASTSURFER_* files)
   2. update the links to examples etc. in all Markdown files (examples moved from README.md to doc/overview/FASTSURFER_EXAMPLES.md)

doc/conf.py
```diff
@@ -21,11 +21,9 @@
 sys.path.append(os.path.dirname(__file__) + "/../recon_surf")
 
 project = "FastSurfer"
-copyright = "2023, Martin Reuter"
-author = "Martin Reuter"
+copyright = "2024, FastSurfer Team"
+author = "FastSurfer Team"
 copyright = f"{date.today().year}, {author}"
```
engrosamaali91:
```suggestion
author = "FastSurfer Developers"
copyright = f"2020, {author}"
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,63 @@
...
+        shell: bash
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: ./dev
+      - name: Setup Python 3.10
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.10
```
m-reuter: probably need to be : ```python-version: '3.10' ```

dkuegler:
```suggestion
          python-version: '3.10'
```

dkuegler (changes requested):
As stated in PR.

pyproject.toml
```diff
@@ -47,7 +47,28 @@ dependencies = [
     'torchio>=0.18.83',
     'tqdm>=4.64',
 ]
-
+doc = [
```
m-reuter, line 51:
```suggestion
[project.optional-dependencies]
doc = [
```

pyproject.toml
```diff
@@ -47,7 +47,28 @@ dependencies = [
...
+    'pypandoc',
+    'nbsphinx',
+    'IPython', # For syntax highlighting in notebooks
+    'ipykernel',
+]
+all = [
+    'fastsurfer[doc]'
+]
+full = [
+    'fastSurfer[all]',
```
m-reuter:
```suggestion
    'fastsurfer[all]',
```

dkuegler (approved):

pyproject.toml
```diff
@@ -47,9 +47,38 @@ dependencies = [
...
+    'sphinx-design',
+    'sphinx-gallery',
+    'sphinx-issues',
+    'pypandoc',
+    'nbsphinx',
+    'IPython', # For syntax highlighting in notebooks
+    'ipykernel',
+    'scikit-image',
+    'torchvision',
+    'sklearn',
```
dkuegler:
```suggestion
    'scikit-learn',
```

dkuegler:
We are finally at a stage where the build works and the error is because of something completely different. In this case, the download_checkpoints.py script is broken, because it was not updated by #467 .

m-reuter:
What still needs to be done here: 

- [x] Import of download_checkpoints fails as it gets `ImportError: cannot import name 'VINN_AXI' from 'FastSurferCNN.utils.checkpoint'` and other similar things. This needs to be fixed in dev (@agirodi ), then we rebase here.
- [x] doc/overview/FASTSURFER_EXAMPLES.md etc still exist and seem to be duplicates, check, merge (if necessary) and delete duplicates. FASTSURFER_EXAMPLES should be deleted , EXAMPLES.md kept, same with the other two.
- [ ] also doc/overview/README.md and the other md files replicate content of /README.md? So this means updating both files all the time? We need to discuss a solution for this. 
- [x] directory /doc/_static needs to be created (with `.gitkeep` empty file in it) as sphinx is looking for it. 
- [x] no idea what to do with the futures formatting warnings: ```concurrent.futures._base.Executor.map:15: WARNING: Inline emphasis start-string without end-string.``` , anyone? Why is sphinx even looking at those as they are not part of our project?

dkuegler:
>     * [ ]  no idea what to do with the futures formatting warnings: `concurrent.futures._base.Executor.map:15: WARNING: Inline emphasis start-string without end-string.` , anyone? Why is sphinx even looking at those as they are not part of our project?

That is even a private module of `concurrent.futures`.

dkuegler:
Maybe we can fix this warning with the autodoc_mock_imports directive?
> Add a list of modules in the [autodoc_mock_imports](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports) to prevent import errors to halt the building process when some external dependencies are not importable at build time.

FastSurferCNN/utils/common.py
```diff
@@ -1075,7 +1074,7 @@ def is_file(p: Path):
             return all(pool.map(is_file, self._subjects))
 
 
-class SerialExecutor(Executor):
+class SerialExecutor:
```
dkuegler:
```suggestion
class SerialExecutor(Executor):
```
Revert inheritance

m-reuter (approved):

PR #476, opened 2024-03-08
Fix python call to spherical_project

dkuegler:
This Commit fixes three issues:
1. tcsh does some weird argument splitting splits variables that should not be split, specifically: tcsh ... --py 'python3.10 -s' is split into "--py", "'python3.10", and "-s'". To fix this, the fs_time script was ported to bash, as it was impossible to accomplish this fix in tcsh.
2. spherical_project always failed, because lapy imports were outdated (lapy.diffGeo instead if lapy.diffgeo and lapy.triaMesh.TriaMesh instead of lapy.TriaMesh). This was a silent fail, as stderr was discarded in spherical_project_wrapper.py and so the fallback was always used.
3. spherical_project further "often" failed inside docker, when no user name could be found by getpass.getuser (in nibabel.io.freesurfer.write_geometry) -- raising an exception. (Happens if the user is defined for docker via -u, but there is no associated username inside the docker.) To fix this, we just add the environment variable "USERNAME" with value "UNKNOWN", if it is not set (this is a place getpass looks for the username last).

recon_surf/functions.sh
- proper quoting of cmd 

recon_surf/spherical_project.py
- fix lapy import

recon_surf/spherical_project_wrapper.py
- also print stderr output to the console
- call spherical_project with USERNAME set 

recon_surf/fs_time
- port to bash

### DIFF
--- a/recon_surf/fs_time
+++ b/recon_surf/fs_time
@@ -38,8 +38,7 @@ If the env variable FSTIME_LOAD is set to 1 or not set at all, then
 uptime is run before and after each process to give the load on the
 system (see below for output)
 
-Default fs_time Output
-} (see also the manual page for /usr/bin/time):
+Default fs_time Output (see also the manual page for /usr/bin/time):
 1. Key (default is @#@FSTIME)
 2. Time stamp at the onset of execution
 3. Command name
--- a/recon_surf/fs_time
+++ b/recon_surf/fs_time
@@ -1,152 +1,33 @@
-#!/bin/tcsh -f
+#!/bin/bash
 # fs_time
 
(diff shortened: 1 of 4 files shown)


m-reuter:
thanks. do you know if this has been an issue already in the last release? Probably it also falls back to FreeSurfer in all cases then?

dkuegler:
It should have been an issue since the lapy API was updated. I don't know when this was.
But theoretically, there are hints in the validation logs.

dkuegler:
Correction 1 was introduced by my last PR... 2/3 would have caused failures before, but I don't know since when exactly

PR #480, opened 2024-03-11
The version script had a hard-coded flag to never use git

dkuegler:
This forced the build_script to not use git in its build script to generate a new BUILD.info file. Which in turn meant that the build script tried to read from the BUILD.info file to create a new BUILD.info file. This lead to a circular dependency and a crash during build.

This PR fixes this issue by:
- removing the hard-coded git flag
- adding an option to the version script to NOT read the BUILD.info file

### DIFF
--- a/FastSurferCNN/version.py
+++ b/FastSurferCNN/version.py
@@ -253,7 +253,7 @@ def main(
     build_cache : False, TextIOWrapper, optional
         A file-like object to read cached version information, the format should be
         formatted like the output of `main`. Defaults to $PROJECT_ROOT/BUILD.info.
-        If build_cache is false, it is ignored.
+        If build_cache is False, it is ignored.
     file : TextIOWrapper, optional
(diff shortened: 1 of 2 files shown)


FastSurferCNN/version.py
```diff
@@ -238,9 +250,10 @@ def main(
     project_file : TextIOWrapper, optional
         A file-like object to read the projects toml file, with the '[project]' section
         with a 'version' attribute. Defaults to $PROJECT_ROOT/pyproject.toml.
-    build_cache : TextIOWrapper, optional
+    build_cache : False, TextIOWrapper, optional
         A file-like object to read cached version information, the format should be
         formatted like the output of `main`. Defaults to $PROJECT_ROOT/BUILD.info.
+        If build_cache is false, it is ignored.
```
dkuegler:
```suggestion
        If build_cache is False, it is ignored.
```

PR #487, opened 2024-03-16
Update install_env.py

jrussell9000:
Adding periods and underscores to the package_pattern to allow Python packages with those characters (allowable in pypi) to be added to environment YAMLs.


dkuegler (commented):
I did not see this was already merged :(

Issue #486, opened 2024-03-16
Use of '--rm' argument in build.py results in a TypeError


jrussell9000:
## Description
Use of '--rm' argument when running docker build script results in TypeError.

## Steps to Reproduce
Clone dev branch, 'python build.py --device cu118 --tag fast:dev --target runtime --rm'.

![image](https://github.com/Deep-MI/FastSurfer/assets/25013843/a4b63e28-16f6-49fe-8cc4-e6d3789e48c8)

## Environment
 - FastSurfer Version: Current dev branch as of 03162024
 - FreeSurfer Version: N/A
 - OS: Ubuntu 22.04
 - CPU: Intel Xeon 8164
 - GPU: Tesla V-100
 - Python 3.10.13 in a conda environment

dkuegler:
Thanks for reporting this issue.

I think we will just remove support for the `--rm` flag. This flag is discouraged, because it means that the time-consuming steps of the build will not be cached, which makes little sense... So the benefit of this flag is not significant :).

jrussell9000:
Works for me!

m-reuter:
I think we should keep this open until the --rm flag is removed or solved. Also I had other issues with the build script which should be fixed at the same time.

dkuegler:
Both these issues should be addressed in #489 .

#489 removes the --rm flag and tries to allow building without buildx... I cannot really check that as all our machines seem to have buildx.

PR #489, opened 2024-03-18
Fix save paths of checkpoints, fix download_checkpoints script

dkuegler:
Fix save paths of checkpoints
fix download_checkpoints script
Consistently introduce the Plane Type to FastSurferCNN.utils 
Some formatting fixes across multiple files
fix extract_recon_surf_time_info.py


dkuegler:
Also addresses #486 and updates to #488.

PR #493, opened 2024-03-22
add code-style workflow

m-reuter:


m-reuter:
I will merge this now. That way everyone can start fixing some of these errors while they work on their files anyway. Once the next biggest PRs are merged, we will clean up everything systematically.

PR #498, opened 2024-03-30
Fix format specifier in find_device memory requirement check

Pmoonesi:
There was a small error in `utils/common.py` that led to a **ValueError** when attempting to print the required memory. The format specifier was missing precision, resulting in the error. I've included screenshots of the error and how it was resolved after correcting the format specifier.

PS: I'm using the latest FastSurfer docker image.

Command:
```bash
docker run --gpus all  deepmi/fastsurfer:latest --fs_license /fs_license/license.txt --t1 /data/structural.nii --sid 003_S_0907 --sd output --parallel
```

The error:
error

After resolution:
resolved


dkuegler (approved):

dkuegler:
Thanks for providing this fix!

Issue #505, opened 2024-04-08
conform.py bug


neurolabusc:
## Description

The [conform.py](https://github.com/Deep-MI/FastSurfer/blob/5cd0b2cd1f5a63eda65aa19a7e5867b025c407d0/FastSurferCNN/data_loader/conform.py#L283) has a bug, which will impact behavior whenever the parameter f_low is not zero. Fortunately, zero is the default for f_low. Fixing this would allow sophisticated usage and development.

The issue is that src_min is initially set to the global minimum. However, f_low is subsequently updated to the lower output threshold. The problem is that final computation of the output high threshold (src_max) assumes the value of src_min is still the global minimum:
```
    src_min = idx * bin_size + src_min

    # get upper limit
    nth = voxnum - int((1.0 - f_high) * nz)
    idx = np.where(cs >= nth)

    if len(idx[0]) > 0:
        idx = idx[0][0] - 2

    else:
        print("ERROR: rescale upper bound not found")

    src_max = idx * bin_size + src_min
```
An easy fix would be to explicitly create and refer to a global_min variable that is independent from the output minimum:
```
    global_min = src_min
    src_min = idx * bin_size + global_min

    # get upper limit
    nth = voxnum - int((1.0 - f_high) * nz)
    idx = np.where(cs >= nth)

    if len(idx[0]) > 0:
        idx = idx[0][0] - 2

    else:
        print("ERROR: rescale upper bound not found")

    src_max = idx * bin_size + global_min
```

dkuegler:
I think I have just confirmed the existence of this bug. When we fix this, we might as well make the math easier to understand though.

dkuegler:
This has been fixed by #506.

PR #506, opened 2024-04-08
Fix getscale for f_low != 0 (default value)

dkuegler:
- As reported in #505 , getscale unexpectedly corrupts src_max (via src_min), if f_low is not 0.0 (the default value).
- Updated variable names, formatting, comments
- Optimized double calls to min and max (explicitly getscale and implicitly via np.histogram).

TODO:
- Needs testing, i.e. test whether results for current and old implementation are the same based on `mri_diff` evaluation.


dkuegler:
Tested and same output confirmed

Issue #518, opened 2024-05-03
Model download issue


roohy466:
Hi, Seems the latest version of run_fastsurfer has issue downloading the models:

    [INFO: run_prediction.py:  652]: Checking or downloading default checkpoints ...
    [INFO: checkpoint.py:  341]: Downloading checkpoint aparc_vinn_axial_v2.0.0.pkl from https://b2share.fz-juelich.de/api/files/a423a576-220d-47b0-9e0c-b5b32d45fc59
    [WARNING: checkpoint.py:  351]: Server https://b2share.fz-juelich.de/api/files/a423a576-220d-47b0-9e0c-b5b32d45fc59 not reachable.
    [INFO: checkpoint.py:  341]: Downloading checkpoint aparc_vinn_axial_v2.0.0.pkl from https://zenodo.org/records/10390573/files
    [WARNING: checkpoint.py:  351]: Server https://zenodo.org/records/10390573/files not reachable.
    Traceback (most recent call last):
      File "/usr/local/KUL_apps/FastSurfer/FastSurferCNN/run_prediction.py", line 806, in <module>
        sys.exit(main(**vars(_args)))
      File "/usr/local/KUL_apps/FastSurfer/FastSurferCNN/run_prediction.py", line 656, in main
        get_checkpoints(ckpt_ax, ckpt_cor, ckpt_sag, urls=urls)
      File "/usr/local/KUL_apps/FastSurfer/FastSurferCNN/utils/checkpoint.py", line 397, in get_checkpoints
        check_and_download_ckpts(file, urls)
      File "/usr/local/KUL_apps/FastSurfer/FastSurferCNN/utils/checkpoint.py", line 379, in check_and_download_ckpts
        download_checkpoint(checkpoint_path.name, checkpoint_path, urls)
      File "/usr/local/KUL_apps/FastSurfer/FastSurferCNN/utils/checkpoint.py", line 354, in download_checkpoint
        raise requests.exceptions.RequestException("No server reachable.")
    requests.exceptions.RequestException: No server reachable.
    ERROR: FastSurfer asegdkt segmentation failed.

dkuegler:
Thank you for reporting this issue!

Edit: sorry, I misread the message.
Can you please verify that
1. The problem still persists
2. you are able to reach zenodo.org from where you called Fastsurfer?

We specifically implemented code in #467 and #488 so that we have two servers from which to download the model weights.

dkuegler:
[...]

PR #519, opened 2024-05-03
Update checkpoint.py

dkuegler:
Add additional debug information if the download of checkpoints fails.


dkuegler:
@agirodi please check if this works and does not cause a new bug. This PR is intended to provide additional information in cases such as #518 to better be able to debug download failures.

agirodi (approved):
Works as intended

Issue #528, opened 2024-06-16
recon_surf//functions.sh: line 38: -debug: command not found


constructor-s:
It seems `fs_time` (`$timecmd`) should support a `-debug` flag but when called in the following manner on a native installation on my Mac (M1, zsh 5.9 (x86_64-apple-darwin22.0), GNU bash, version 5.2.26(1)-release (aarch64-apple-darwin22.6.0)), it does not like it. When I removed it it seemed to proceed okay. Not sure if it is an issue with the Mac shell.

https://github.com/Deep-MI/FastSurfer/blob/3622c13c049a367e72ca9c0db8332c0493101db7/recon_surf/functions.sh#L38

```
...

./recon-surf.sh --sid 140 --sd /Users/USERNAME/my_fastsurfer_analysis --t1 /Users/USERNAME/my_fastsurfer_analysis/140/mri/orig.mgz --asegdkt_segfile /Users/USERNAME/my_fastsurfer_analysis/140/mri/aparc.DKTatlas+aseg.deep.mgz --threads 1 --parallel

freesurfer-macOS-darwin_x86_64-7.4.1-20230614-7eb8460
$Id$
Darwin USERNAMEs-MacBook-Pro-2.local 22.6.0 Darwin Kernel Version 22.6.0: Mon Apr 22 20:50:39 PDT 2024; root:xnu-8796.141.3.705.2~1/RELEASE_ARM64_T8103 arm64
 
================== Checking validity of inputs =================================
 
 
 RUNNING both hemis in PARALLEL 
 RUNNING 1 number of OMP THREADS 
 RUNNING 1 number of ITK THREADS 
 
Checking Input Segmentation Quality ...
python3.10 /Users/USERNAME/FastSurfer_dev/FastSurferCNN/quick_qc.py --asegdkt_segfile /Users/USERNAME/my_fastsurfer_analysis/140/mri/aparc.DKTatlas+aseg.deep.mgz
/Users/USERNAME/FastSurfer_dev/recon_surf//functions.sh: line 38: -debug: command not found
```

m-reuter:
Hi, not sure what is going on maybe @dkuegler can take a look as I found this commit: https://github.com/Deep-MI/FastSurfer/commit/56ac5fb998f0d05150f1905c5e5e5393d948c225 which says it adds support for the -debug flag.

dkuegler:
First off, thank you for reporting this issue.

[...]

PR #530, opened 2024-06-19
Improve Mac support and clean up dependencies

m-reuter:
On the mac we only have older bash versions. Installing a newer bash 4 or 5 via brew is possible, but sub-scripts still include /bin/bash via the # first line directive (reported here #528 ). Modifying that line could break portability to other OSs (maybe ```#!/usr/bin/env bash``` is portable but poses a privilege escalation security threat). Therefore we replace all bash 4.* functionality with alternatives that are supported with older bash 3.1.

- ```|&```gets replaced with ```2>&1 |```
- ```@Q``` expansion gets replaced with ```printf %q```(available since bash 3.1)
- requirememnts.mac.txt gets updated to newer dependency versions (no more torch dev required etc) in accordance to updated requirements.txt. It also now allows version ranges, making it less repeatable, but more flexible. 
- pyproject.toml now includes some missing dependencies and also bumps versions in accordance with requirements.txt
- INSTALL.md text gets updated accordingly for mac native install (no newer bash needed)
- deprecated panda ```delim_whitespace?``` gets updated to ```sep='\s+'``` (this is really independent of the mac version).


requirements.mac.txt
```diff
@@ -1,185 +1,18 @@
...
+pandas>=1.5.3
+pyyaml>=6.0
+scikit-image>=0.19.3
+scikit-learn>=1.2.2
+scipy>=1.10.1,!=1.13.0
+simpleitk>=2.2.1
+tensorboard>=2.12.1
+torch>=2.0.1+cu117
+torchio>=0.18.83
+torchvision>=0.15.2+cu117
```
dkuegler: The cuXXX designations should not work on mac. I think you need to remove that.

dkuegler (commented):
All else seems fine.

requirements.mac.txt
m-reuter: it worked, but I removed it anyway. thanks.

PR #537, opened 2024-06-27
Fixes of HypVINN & options for conform that disable critieria

dkuegler:
- Some fixes to the HypVINN pipeline 
  - formatting/doc
  - naming of HypoVINN->HypVINN (unified)
  - fix some errors with default hypvinn file names and arguments
  - generally improve and clean up the hypvinn interface and functions
- conform.py gets new flags `--no_lia`, `--no_strict_lia`, `--no_img_size`, `--no_vox_size` which deactivate parts of the conform script:
  - `no_lia`: rotation of the affine is not touched, image may remain RAS
  - `no_strict_lia`: rotation of the affine is adapted, but only towards LIA such that no interpolation would be necessary, but will also be LIA (never RAS)
  - `no_img_size` no padding/cropping of the image
  - `no_iso_vox`~`no_vox_size`~ no harmonization of voxel size, for example in combination with `no_strict_lia`, the image is not made iso. voxelsize
-  Cleanup of shell scripts
  -  Remove extra "2&>1" in shell scripts in echo statements (they never have stderr output)

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -1,4 +1,5 @@
-# Copyright 2019 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
+# Copyright 2019
+# AI in Medical Imaging, German Center for Neurodegenerative Diseases (DZNE), Bonn
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,12 +15,14 @@
 
 
 # IMPORTS
-import logging
-from typing import Optional, Type, Tuple, Union, Iterable, cast
 import argparse
+from enum import Enum
+import logging
 import sys
+from typing import Optional, Type, Tuple, Union, Iterable, cast
 
 import numpy as np
+import numpy.typing as npt
 import nibabel as nib
 
 from FastSurferCNN.utils.arg_types import (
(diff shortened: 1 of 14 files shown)


FastSurferCNN/data_loader/conform.py
```diff
@@ -107,16 +129,42 @@ def options_parse():
...
+    )
+    advanced.add_argument(
+        "--no_lia",
+        dest="force_lia",
+        action="store_false",
+        help="Ignore the reordering of data into lia (without interpolation). "
+             "Superceeds --no_strict_lia",
+    )
+    advanced.add_argument(
+        "--no_vox_size",
```
m-reuter: maybe better no_iso_vox ?? or no_vox_iso ?

FastSurferCNN/data_loader/conform.py
```diff
@@ -107,16 +129,42 @@ def options_parse():
...
-        help="Specifies the target data type of the target image or 'any' (default: 'uint8', "
-        "as in FreeSurfer)",
+        help="Specifies the target data type of the target image or 'any' (default: "
+             "'uint8', as in FreeSurfer)",
+    )
+    advanced.add_argument(
+        "--no_strict_lia",
+        dest="force_strict_lia",
+        action="store_false",
+        help="Ignore the forced lia reorientation.",
```
m-reuter: use "LIA" instead of "lia" also in line 154.

FastSurferCNN/data_loader/conform.py
```diff
@@ -107,16 +129,42 @@ def options_parse():
...
+    advanced.add_argument(
+        "--no_lia",
+        dest="force_lia",
+        action="store_false",
+        help="Ignore the reordering of data into lia (without interpolation). "
+             "Superceeds --no_strict_lia",
+    )
+    advanced.add_argument(
+        "--no_vox_size",
+        dest="force_vox_size",
```
m-reuter: also here, maybe better force_iso_vox

FastSurferCNN/data_loader/conform.py
dkuegler: I guess that is also true and way easier to understand

FastSurferCNN/data_loader/conform.py
```diff
@@ -107,16 +129,42 @@ def options_parse():
...
+        action="store_false",
+        help="Ignore the forced isometric voxel size (depends on --conform_min).",
+    )
+    advanced.add_argument(
+        "--no_img_size",
```
m-reuter, line 164: why not no_img_dim ?

dkuegler: dim is usually the number of dimensions, so usually 3 or 4 in python (numpy/torch)
shape would be what we mean, but I am not sure it that is more "helpful"

FastSurferCNN/data_loader/conform.py
```diff
@@ -560,33 +661,57 @@ def conform(
...
+        The epsilon for the affine check.
+
+    Returns
+    -------
+    bool
+        The result.
+    """
+    _v2v = np.abs(vox2vox[:3, :3])
+    # check 1: have exactly 3 times 1/-1 rest 0, check 2: all 1/-1 or 0
+    return abs(_v2v.sum() - 3) > eps or np.any(np.maximum(_v2v, abs(_v2v - 1)) > eps)
```
m-reuter, line 701: shouldn't this be abs(_v2v.sum()) - 3 ? and also abs(_v3v) - 1 ?

dkuegler: v2v is already abs in line 699

recon_surf/recon-surf.sh
```diff
@@ -531,16 +531,16 @@ if [[ ! -f "$mdir/transforms/talairach.lta" ]] || [[ ! -f "$mdir/transforms/tala
   echo " " 2>&1 | tee -a $LF
   echo "============= Computing Talairach Transform ============" 2>&1 | tee -a $LF
   echo " " 2>&1 | tee -a $LF
```
m-reuter: Here (and maybe other places) the 2>&1 can also be removed

FastSurferCNN/data_loader/conform.py
m-reuter: yes, I think size is fine

PR #541, opened 2024-07-09
Fix MGHheader set data type attribute error

santiestrada32:
I found an error in the name of the function that sets the data type on the MGHHeader.


dkuegler (approved):
Yeah, this is a recent change. Good fix!

PR #547, opened 2024-07-16
Fix docker build script, only works with docker-container builder, if buildx is installed

dkuegler:
Fix checks for the buildx/docker-container requirements. Used to always require docker-container builder, if buildx was installed. This requirement is a bit too strict. The docker builder is good enough, if no attestation or caching are selected.


dkuegler:
Should probably be merged into dev soon-ish. It fixes some smaller issues with the docker build scripts.

PR #552, opened 2024-08-06
Separate doc for stable and remove doctrees folder

dkuegler:
Add stable targets as an additional target
Remove doctree files from deployment: doctree files are binary files bloating the gh_pages branch unnecessarily.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -5,7 +5,7 @@ concurrency:
 on:
   pull_request:
   push:
-    branches: [dev]
+    branches: [dev, stable]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


dkuegler:
I would combine accepting this PR with prior to this resetting the gh_pages branch to the original state.

One question that might be open is: how does the multiple folders and action on stable and dev interact with the gh_pages branch. I hope the dev folder does not get deleted when the stable docs are updated. I cannot easily test this though...

dkuegler:
This PR does not pass checks because sphinx 8 somehow treats importing of files different ot sphinx 7. This needs to be debugged at some point, but we should just use sphinx 7 for now.

PR #553, opened 2024-08-07
Small fixes to shell scripts

dkuegler:
Add a checking for subjects_dir to run_fastsurfer.sh (whether subject_dir is set, is root, is writable)
Bump tqdm to 4.66 (has no effect on results)
add BATCH documentation (brun_fastsurfer.sh usage docs and TQDM_DISABLE=1 doc) Add --env TQDM_DISABLE=1 for slurm running to pollute the logs less Fix the parsing of subjects in brun_fastsurfer.sh
Clean up the parameter parsing and fix some minor Shellshock Warning in srun_fastsurfer.sh

### DIFF
--- a/brun_fastsurfer.sh
+++ b/brun_fastsurfer.sh
@@ -128,8 +128,7 @@ case $key in
     ;;
   --subjects)
     subjects_stdin="false"
-    while [[ "$1" =~ ^-- ]] ; do subjects=("${subjects[@]}" "$key") ; shift
-    done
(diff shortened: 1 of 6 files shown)


dkuegler:
This PR does not pass checks because sphinx 8 somehow treats importing of files different to sphinx 7. This needs to be debugged at some point, but we should just use sphinx 7 for now.

The limitation to sphinx 7 is implemented in #552, so this PR's checks should rerun and be merged after merging #552.

PR #554, opened 2024-08-07
Fallback for unoriented triangles in sample_parc

ClePol:
This fallback fixes rare crashes of the recon_surf pipeline, where mris_place_surface generates triangle meshes with inconsistent orientation. The added check detects and fixes these cases. I recommend to include the failing cases in the release testing for the next release to make sure there are no other effects somewhere else in the pipeline.

### DIFF
--- a/recon_surf/rewrite_oriented_surface.py
+++ b/recon_surf/rewrite_oriented_surface.py
@@ -1,0 +1,102 @@
+# Copyright 2024 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# IMPORTS
+import sys
+import argparse
+from pathlib import Path
+
+import lapy
+
+__version__ = "1.0"
+
+
+def make_parser() -> argparse.ArgumentParser:
+    """
+    Create a command line interface and return command line options.
+
+    Returns
+    -------
+    options
+        Namespace object holding options.
+    """
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


dkuegler:
This PR does not pass checks because sphinx 8 somehow treats importing of files different to sphinx 7. This needs to be debugged at some point, but we should just use sphinx 7 for now.

The limitation to sphinx 7 is implemented in #552, so this PR's checks should rerun and be merged after merging #552.

dkuegler (approved):
Looks good to me :)

dkuegler:
I added a new rewrite_oriented_surface script and integrated that into recon-surf. going to run the test cases again as well to see if it fixes things.

dkuegler:
Currently awaiting testing.

recon_surf/rewrite_oriented_surface.py
```diff
@@ -0,0 +1,89 @@
...
+# limitations under the License.
+
+# IMPORTS
+import sys
+import argparse
+from pathlib import Path
+
+import lapy
+
+__versoion__ = "$Id: rewrite_oriented_surface,v 1.0 2024/08/07 17:20:10 kueglerd $"
```
m-reuter: typo in version

m-reuter:
currently you fix the surface after the second inflation step, but probably the first time the surface has flipped triangles is earlier? Do you know in what exact freesurfer commend (mris_make_surfaces inside what recon-all step) this happens? We should probably fix it directly there.

dkuegler:
So the last version crashed, because nibabel write_geometry uses getpass.getuser, which crashes, if it cannot find a username, which happens if run in a docker container. I hope I fixed this issue now. Tests are running.

> currently you fix the surface after the second inflation step, but probably the first time the surface has flipped triangles is earlier? Do you know in what exact freesurfer commend (mris_make_surfaces inside what recon-all step) this happens? We should probably fix it directly there.

So I just put it after the call to recon-all that creates the surface, to follow your proposal, we would have to split the different stages to the recon-all call.

m-reuter:
The previous recon all call: 
```cmd="recon-all -subject $subject -hemi $hemi -smooth2 -inflate2 -curvHK -no-isrunning $hiresflag $fsthreads"```
does not create the surfaces, that happens further up, but even there we might have to split the call. The question is, do we even know where exactly the problem starts (at creation of the first surface, at moving/placing the surfaces to WM or Pial) etc?

dkuegler:
I have tested this version now and the results are as they should be. Files generated as intended and the volume information is fixed, so lapy updates are not required for proper function.

PR #559, opened 2024-08-22
replace .A and .A1 with explicit array conversion

m-reuter:
Newer versions of SciPy seem to have forgotten the short functions .A() and .A1() for array conversion, so we call explicitly the long names.

### DIFF
--- a/recon_surf/smooth_aparc.py
+++ b/recon_surf/smooth_aparc.py
@@ -212,8 +212,8 @@ def mode_filter(
     # get rid of rows that have uniform vote (or are empty)
     # for this to work no negative numbers should exist
     # get row counts, max and sums
-    rmax = nlabels.max(1).A.squeeze()
-    sums = nlabels.sum(axis=1).A1
+    rmax = nlabels.max(1).toarray().squeeze()
+    sums = nlabels.sum(axis=1).toarray().ravel()
     counts = np.diff(nlabels.indptr)
     # then keep rows where max*counts differs from sums
     rmax = np.multiply(rmax, counts)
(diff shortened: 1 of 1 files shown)


dkuegler:
@taha-abdullah please test this with scipy 1.14.1

taha-abdullah:
I am getting the following error when I run this:

Traceback (most recent call last):
  File "/fastsurfer/recon_surf/sample_parc.py", line 430, in <module>
    sample_parc(options.insurf, options.inseg, options.seglut, options.surflut, options.outaparc, options.incort, options.projmm, options.radius)
  File "/fastsurfer/recon_surf/sample_parc.py", line 421, in sample_parc
    smooths = smooth_aparc(surf, surfsamples, cortex)
  File "/fastsurfer/recon_surf/smooth_aparc.py", line 335, in smooth_aparc
    labels_new = mode_filter(adjM, labels, fillonlylabel, np.array([fillonlylabel]))
  File "/fastsurfer/recon_surf/smooth_aparc.py", line 216, in mode_filter
    sums = nlabels.sum(axis=1).toarray().squeeze()
AttributeError: 'matrix' object has no attribute 'toarray'
@#@FSTIME  2024:08:23:11:00:19 recon_surf/sample_parc.py N 16 e 0.95 S 0.12 U 0.83 P 100% M 189276 F 0 R 51683 W 0 c 11 w 1 I 0 O 0 L 3.12 3.06 3.31
@#@FSLOADPOST 2024:08:23:11:00:20 recon_surf/sample_parc.py N 16 3.03 3.04 3.30

dkuegler:
This error is with scipy 1.14.1 u
I assume?

m-reuter:
I assumed that sum produces a sparse matrix just as max, but it produces a matrix, so array conversion is slightly different. I also fixed deprecated in1d. These changes have now been tested with numpy 2.1.0 and script 1.14.1 and yield identical results to the previous version in my test case.

PR #560, opened 2024-08-23
update versions of python packages

dkuegler:
This PR updates the versions and code of python packages and cuda in the docker build environment to most recent versions.
Some warnings because of deprecation or unsafe behaviors (that should not affect FastSurfer, because we do not download and use unsafe checkpoints) are also fixed.

This PR is rebased on #559.

Validation testing is running for this version (admittedly on the version before the fix to the torch warning (last commit)).


Docker/build.py
```diff
@@ -61,8 +60,8 @@ class DEFAULTS:
     # torch 2.0.1 comes compiled with cu117, cu118, and rocm5.4.2
```
m-reuter, line 60: comment needs update.

m-reuter:
and we need to test if the new rocm works.

m-reuter:
More questions:
- requirements.txt has not been updated accordingly
- are all packages on the newest versions? for example numpy is still not 2.* (of course we can postpone that, but should document here why some packages are not bumped and what needs to be done to do that later).

m-reuter:
Problems in the doc build: 

```
reading sources... [ 99%] scripts/segstats
:1: (WARNING/2) Inline literal start-string without end-string.
```

and 

```
WARNING: 'myst' cross-reference target not found: 'README.md'
```

in a few md files.

dkuegler:
This PR currently only needs an update of the requirements files, which I currently cannot generate.

PR #573, opened 2024-09-03
Add code style via ruff

m-reuter:
This switches on our code-style workflow mainly to run Ruff and spellchecker. So this fixes ruff messages in all code (lots of changes, but mainly very minor stuff).


Docker/install_fs_pruned.sh
```diff
@@ -97,7 +97,7 @@ function run_parallel ()
 }
 
 
-# get Freesurfer and upack (some of it)
+# get FreeSurfer and upnack (some of it)
```
dkuegler:
```suggestion
# get FreeSurfer and unpack (some of it)
```

PR #578, opened 2024-09-16
Added explicit creation of nonroot user to Dockerfile

jrussell9000:
Added explicit creation of non-root to the dockerfile. In some situations (e.g., containerd on Kubernetes) the home directory for the user (/home/nonroot) isn't necessarily created when the image is loaded. In this case, things get tricky because the nonroot user doesn't seem to have write access to any other paths. Also, explicitly coding in the useradd seems to be best practice (though this is obviously less relevant).

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -221,6 +221,7 @@ RUN cd /fastsurfer ; python3 FastSurferCNN/download_checkpoints.py --all && \
 
 # Set FastSurfer workdir and entrypoint
 #  the script entrypoint ensures that our conda env is active
+RUN useradd -m -s /bin/bash -u 1000 -g 1000 nonroot
 USER nonroot
 WORKDIR "/fastsurfer"
 ENTRYPOINT ["/fastsurfer/Docker/entrypoint.sh","/fastsurfer/run_fastsurfer.sh"]


dkuegler:
Hi jrussell9000,
can you please explain how FastSurfer is causing problems with kubernetes?

Thus far, it has been a design decision to not create the nonroot user inside of the container. This is so users are forced to pass the `-u $(id -u):$(id -g)` flags into the docker container upon creation (`docker run -u $(id -u):$(id -g) --rm ...`). Otherwise your output files may belong to a random user in your file system (on the host computer) and that is not really great.

Unless there is a very specific reason why you need the nonroot user to be existing, I would rather not have it exist. 

As a stop-gap solution for you, you can just create/build your own image based on the fastsurfer image like this:
```bash
docker build --tag myfastsurfer <<EOF
from deepmi/fastsurfer:cu124-v2.3.0
USER root
RUN useradd -m -s /bin/bash -u 1000 -g 1000 nonroot
USER nonroot
EOF
``` 

This should generate you an image that has what you want. Going a step further, you can even insert your own user name and user id and group id (and copy your FreeSurfer license, etc.).
Cheers

jrussell9000:
That's an entirely reasonable explanation. Briefly described, I'm running FastSurfer in the context of an Argo Workflow (on a Kubernetes cluster). The FastSurfer step halts during container setup with,  'CreateContainerError: failed to create containerd container: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount1150146980: no users found'. I've tried explicitly running as nonroot (i.e., uid 1000) but no luck.  No luck Googling this so far, either. The closest I came _seemed_  to suggest that the issue might be solved by explicitly adding the user during image creation. 

I'll consider creating my own image - I was hoping to avoid adding an extra step - and go bug the Argo folks for more help.  Either way, I'll withdraw the pull.  Thanks!

PR #583, opened 2024-10-01
Long recon-surf

m-reuter:
Initial modifications to recon-surf.sh for longitudinal processing

### DIFF
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -1183,6 +1183,22 @@ if [ "$base" != "1" ] ; then
            --threads "$threads")
 #      cmd="$python $FASTSURFER_HOME/FastSurferCNN/mri_segstats.py --seed 1234 --seg $mdir/wmparc.mgz --sum $statsdir/wmparc.stats --pv $mdir/norm.mgz --in-intensity-name norm --in-intensity-units MR --subject $subject --surf-wm-vol --ctab $FREESURFER_HOME/WMParcStatsLUT.txt --etiv"
     else
+
+      if [ "$long" == "1" ] ; then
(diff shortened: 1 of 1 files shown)


recon_surf/recon-surf.sh
```diff
@@ -31,6 +31,9 @@ threads="1"           # number of threads to use for running FastSurfer
 allow_root=""         # flag for allowing execution as root user
 atlas3T="false"       # flag to use/do not use the 3t atlas for talairach registration/etiv
 segstats_legacy="false" # flag to enable segstats legacy mode
+base=0                # flag for longitudinal template (base) run
+long=0                # flag for longitudinal time point run
+baseid=""             # baseid for logitudinal time point run
```
dkuegler, line 36: Is this not redundant information? I think I would have consolidated long and baseid into one variable, because `baseid!=""` is essentially the same as `long==1`

recon_surf/recon-surf.sh
```diff
@@ -252,6 +261,23 @@ then
...
+    echo "the base (template) has been created when running with --long flag."
+    exit 1
+  fi
+  if [ ! grep -Fxq $basedir/base-tps.fastsurfer $subject ] ; then
+    echo "$subject id not found in base-tps.fastsurfer. Please ensure that this time point"
+    echo "was included during creation of the base (template)."
+    exit 1
+  fi
+fi
+
```
dkuegler, line 280: We also probably need some kind of check to see if this is both a long and a base run (which seems to be invalid)

m-reuter: yes, XOR check would be needed

recon_surf/recon-surf.sh
m-reuter: I don't really care. I thought about it and found that this way long is more parallel to base (boolean) and think it does not add too much overhead. So I would keep it for now.

PR #588, opened 2024-10-04
Longitudinal processing

m-reuter:
This PR adds scripts for template creation and running both template (base) and long. 
Entry script is long_fastsurfer.sh that runs everything sequentially.


m-reuter:
I successfully ran this on a case from OASIS (1mm isotropic, 2 time points). Other resolutions and time points still need testing. 

The script long_fastsurfer.sh is running a full case with all time points sequentially. The steps are:

1. setup of base/template with long_prepare_template.sh
2. running of template through run-fastsurfer.sh --base ...
3. running of each time point through run-fastsurfer.sh --long ...

2 and 3 can be split into seg_only and surf_only. Also the segmentation of the longitudinal time points in 3 can principally be run in parallel with step 2, only the surface module in 3 needs to wait until the base (step 2) is completely done.

PR #591, opened 2024-10-10
Support for edits

dkuegler:
This PR adds support for edits in FastSurfer.

Edits are enabled by the `--edits` flag for `run_fastsurfer.sh`.

Edits supported:
<asegdkt_segfile> via <asegdkt_segfile-w/o-ext>.manedit.<ext>
white matter segmentation (wm.mgz) via wm.manedit.mgz
white matter control points (via the standard FreeSurfer control points file)
talairach registration via mri/mri/transforms/talairach.xfm (overwriting talairach registration via mri/mri/transforms/talairach.xfm)
brain.finalsurfs.mgz via brain.finalsurfs.manedit.mgz

This PR also move the allow_root check into functions.sh and introduces run_it and run_it_cmdf function variants that are more clear wrt. argument splitting (see SC2206, SC2207 and similar).

### DIFF
--- a/doc/overview/EDITING.md
+++ b/doc/overview/EDITING.md
@@ -1,11 +1,17 @@
 # Manual Edits
 
 We have noticed that FastSurfer segmentations and surface results are very robust and rarely require any manual edits. 
-However, for your convenience, we allow manual edits in various stages of the FastSurfer pipeline.
+However, for your convenience, we allow manual edits in various stages of the FastSurfer pipeline to fix errors or inaccuracies in FastSurfer results.
 These editing options include approaches that are inherited from FreeSurfer as well as some FastSurfer-specific editing options.
 
 The provided editing options may be changed or extended in the future, also depending on requests from the community.
 Furthermore, we invite users to [contribute](CONTRIBUTING.md) such changes and/or datasets of paired MRI images and edited files to improve FastSurfer's neural networks.  
+
+## What are Edits?
+
+Edits are manual interventions into the pipeline that change intermediate results.
+By "editing" intermediate results, later steps operate on updated information.
+While rarely necessary, some errors of the pipeline (see below) can be addressed and the quality of results can be improved.
 
 ## What can be edited?
 
@@ -43,9 +49,12 @@ FastSurfer supports the following edits:
 Note, as FastSurfer's surface pipeline is derived from FreeSurfer, some edit options and corresponding naming schemes are inherited from FreeSurfer.
 
 ## General process
-To use the `--edits` flag with `run_fastsurfer.sh`, you must first update the images in the subject directory that require editing, and then reprocess the updated directory using the --edits option.
+Most editing options require that you have processed the case and you are able to access and modify the files in that image's directory, i.e. `<subject_dir>`.
+Based on a quality inspection of results, segmentations and surfaces, one of the edits (see above) is chosen and performed.
+Then, the case is re-processed using the `--edits` flag, i.e. `run_fastsurfer.sh` is re-run (same FastSurfer command and FastSurfer version as before, but add `--edits`).
+Finally, you would check in another quality inspection, if the quality issues are resolved.
 
-For example (including setup for native processing, see [Examples](EXAMPLES.md) for other processing options):
+For example (including the setup for native processing, see [Examples](EXAMPLES.md) for other processing options):
 
 ```bash
 # Setup FASTSURFER and FREESURFER
@@ -124,9 +133,9 @@ In particular, over- and under-segmentation of the brainmask, gray matter over s
 In specific, such errors are inspected in `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.mgz`, `<subject_dir>/mri/aseg.auto_noCCseg.mgz` and `<subject_dir>/mri/mask.mgz`.
 
 ### What to do
-1.Copy the file `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.mgz` to the same directory with a different name, `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.manedit.mgz`.
-2. Open and edit `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.manedit.mgz` and fix any errors you find.
-3. [Re-run FastSurfer](#general-process) to apply the changes you made and update `<subject_dir>/mri/aseg.auto_noCCseg.mgz` and `<subject_dir>/mri/mask.mgz`.
+1. Copy the file `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.mgz` to the same directory and name it `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.manedit.mgz`.
+2. Open `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.manedit.mgz` (for example using Freeview) and resolve all errors/quality issues.
+3. [Re-run FastSurfer](#general-process) to propagate the changes into other results. Among others, this updates `<subject_dir>/mri/aseg.auto_noCCseg.mgz` and `<subject_dir>/mri/mask.mgz`.
 
 ## Talairach registration
 
@@ -134,8 +143,9 @@ In specific, such errors are inspected in `<subject_dir>/mri/aparc.DKTatlas+aseg
 The estimated total intracranial volume is inaccurate.
 
 ### What to do 
-1. Open and edit `<subject_dir>/mri/transforms/talairach.xfm` and replace the edited version with the old file.
-2. [Re-run FastSurfer](#general-process) to update the eTIV value in all stats files.
+1. Copy `<subject_dir>/mri/transforms/talairach.auto.xfm` to `<subject_dir>/mri/transforms/talairach.xfm`, replace the symlink that is already there.
+2. Edit `<subject_dir>/mri/transforms/talairach.xfm` to have the intended talairach matrix, see the [FreeSurfer tutorial for details](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/Talairach).
+3. [Re-run FastSurfer](#general-process) to update the eTIV value in all stats files.
 
 See also: <https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/Talairach>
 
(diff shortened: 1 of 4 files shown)


m-reuter:
Hi, thanks for the pr. It would be great to fix the spelling errors and also add a readme for edits, which can be pretty short in the first iteration, with a 1-2 sentence description of each possible edit.

dkuegler:
Spelling should be fixed, documentation will need to be a future thing for now.

recon_surf/recon-surf.sh
```diff
@@ -1002,36 +1044,39 @@ for hemi in lh rh ; do
...
   echo "popd > /dev/null" >> "$CMDF"
 
   # Here insert DoT2Pial  later --> if T2pial is not run, need to softlink pial.T1 to pial!
   echo "pushd $sdir > /dev/null" >> "$CMDF"
   softlink_or_copy "$hemi.pial.T1" "$hemi.pial" "$LF" "$CMDF"
-  echo "popd > /dev/null" >> "$CMDF"
 
   # these are run automatically in fs7* recon-all and cannot be called directly without -pial flag (or other t2 flags)
   # they are the same for fsaparc and long
-  echo "pushd $mdir > /dev/null" >> "$CMDF"
```
m-reuter: Have you tested this? I am not sure why it was cd into mri dir, but maybe there was a reason. With your change it is now running from within surf dir which might change things.

recon_surf/talairach-reg.sh
```diff
@@ -55,60 +88,113 @@ else
...
 
 # Since we do not run mri_em_register we sym-link other talairach transform files here
-pushd "$mdir/transforms" > /dev/null || ( echo "ERROR: Could not change to the transforms directory $mdir/transforms!" | tee -a "$LF" && exit 1 )
-  cmd=(ln -sf "talairach.xfm.lta" "talairach_with_skull.lta")
-  RunIt "${cmd[*]}" "$LF"
-  cmd=(ln -sf "talairach.xfm.lta" "talairach.lta")
-  RunIt "${cmd[*]}" "$LF"
+pushd "$mdir/transforms" > /dev/null || ( echo "ERROR: Could not change to the transforms directory $mdir/transforms!" | tee -a "$LF" ; exit 1 )
+  if [[ ! -e talairach_with_skull.lta ]] ; then
+    cmd=(ln -sf talairach.xfm.lta talairach_with_skull.lta)
```
m-reuter: shouldn' these be ln_or_copy ?

run_fastsurfer.sh
```diff
@@ -220,6 +227,9 @@ SURFACE PIPELINE:
                             to exist already in this case.
   --3T                    Use the 3T atlas for talairach registration (gives better
                             etiv estimates for 3T MR images, default: 1.5T atlas).
+  --edits                 Disable the check for existing recon-surf.sh run and enables
+                            edits of mri/wm.mgz and brain.finalsurfs.mgz by manedit
+                            files as well as FreeSurfer-style WM control points.
```
m-reuter: I think it is strange to have the flag --edits in the help text at multiple places. I think it should be a single help block explaining all of its effects in one location.

dkuegler: I guess consolidating this into the first text makes more sense. On hindsight, I agree. I think --no_biasfield also has multiple mentions, so that should also be changed.

recon_surf/recon-surf.sh
dkuegler: I had tested some of these, but this clearly is a bug :-(

recon_surf/talairach-reg.sh
dkuegler: It should, I am not sure why, but I believe this has never changed to `softlink_or_copy`

dkuegler:
This branch is currently a mess of fixes and features for 6 different things. Those need to be separated into individual branches and PRs.

At a minimum, this version seems to work for longitudinal processing (still unclear for edits).

m-reuter:
I rebased this to current dev which disentangles the edits from the other changes. Please check :
- that everything is as wanted and
- no purposeful changes from the earlier commits are undone

Also update your local with a hard reset.

dkuegler:
Maybe there should be a check if the fastsurfer versions are the same between the original run and the rerun? @m-reuter 

If this is just a warning or a message this will likely get ignored.
If this is a failure, it might get more annoying than not.
Solution might be to write a message to the log and let the user confirm it (enter).
But if nothing is attached to stdin, the process proceeds no matter what so only interactive sessions stop (this is how install scripts sometimes solve this)...

m-reuter:
> Maybe there should be a check if the fastsurfer versions are the same between the original run and the rerun?

I think warning message in the logs is fine. It should only stop if versions are not compatible any longer (and it would fail somewhere during processing). Otherwise, if someone edits things, that is changing more than most of our versions.

doc/overview/EDITING.md
```diff
@@ -1,125 +1,213 @@
...
+# Run FastSurfer
+$FASTSURFER_HOME/run_fastsurfer.sh \
+    --sd $SUBJECTS_DIR --sid case_with_edits \
+    --t1 $SUBJECTS_DIR/case_with_edits/mri/orig/001.mgz \
+    --fs_license $FREESURFER_HOME/.license \
+    --edits # more flags as needed, e.g. --parallel --3T --threads 4
+```
+
+Note, a re-run of the segmentation pipeline, as in the command above, should not be harmful, but is only required if the [asegdkt_segfile](#asegdkt_segfile) was edited.
+There, we can usually skip the segmentation step with
```
m-reuter: Maybe you mean: "Therefore, in most cases, we can skip ..."

doc/overview/EDITING.md
```diff
@@ -1,125 +1,213 @@
...
+The manual label 255 indicates a voxel should be included in the white matter and a voxel labeled 1 should not.
+
+See also: <https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/WhiteMatterEdits_freeview>
+
+## Pial surface placement
+
+### When to use
+Over- and/or under-segmentation of the cortical gray matter: voxels that should be gray matter are excluded, or those that should not are included.
+
+This explicitly
```
m-reuter: Sentence is missing.

doc/overview/EDITING.md
```diff
@@ -1,125 +1,213 @@
...
+### When to use
+Over- and/or under-segmentation of the cortical gray matter: voxels that should be gray matter are excluded, or those that should not are included.
+
+This explicitly 
+
+### What to do
+Often, these errors should be fixed in [asegdkt_segfile](#asegdkt_segfile) `<subject_dir>/mri/aparc.DKTatlas+aseg.deep.manedit.mgz`, but if that is not successful:
+1. Open end edit `<subject_dir>/mri/brain.finalsurfs.manedit.mgz` (overwriting values in `<subject_dir>/mri/brain.finalsurfs.mgz`).
+2. [Re-run FastSurfer](#general-process) to fix the pial surface. 
+3.
```
m-reuter: item 3. is missing

doc/overview/EDITING.md
```diff
@@ -1,125 +1,213 @@
...
+## Guide for experienced FreeSurfer users
+
+The following list enumerates editing options available in FreeSurfer and how to achieve similar edits in FastSurfer. 
+
+1. Skull stripping by editing `<subject_dir>/mri/brainmask.mgz` (delta to `brainmask.auto.mgz`) -> [asegdkt_segfile](#asegdkt_segfile)
+2. Talairach registration by editing `<subject_dir>/mri/transforms/talairach.xfm` (delta to `transforms/talairach.xfm`) -> [Talairach Registration](#talairach-registration)
+3. Brain extraction by editing `<subject_dir>/mri/brain.mgz` (delta to `brain.auto.mgz`) -> [asegdkt_segfile](#asegdkt_segfile)
+4. Subcortical segmentation: `<subject_dir>/mri/aseg.presurf.mgz` (delta to `aseg.presurf.auto.mgz`) -> [asegdkt_segfile](#asegdkt_segfile)
+5. CRS seeds: PONS, CC, LH, RH, watershed => seed files (`<subject_dir>/scripts/seed-(pons|cc|lh|rh|ws).crs.man.dat`) -> [asegdkt_segfile](#asegdkt_segfile)
+6. WM control points: `<subject_dir>/tmp/control.dat` -> [asegdkt_segfile](#asegdkt_segfile)
```
m-reuter, line 196: hm, we could allow them to create control points which should be used by recon-surf when crating wm.mgz?

doc/overview/EDITING.md
dkuegler: I think you actually wrote this sentence :).

```suggestion
Therefore, in most cases, we can skip the segmentation step with
```

doc/overview/EDITING.md
dkuegler: I think we could, but there are some comments somewhere why we would not want to do that. I think I did not really understand those comment but just assumed you had decided against this on principle.

doc/overview/EDITING.md
dkuegler: I think this should just be removed... The docs here does not really have a step, it would just be to check if the results are now better :)
```suggestion
```

doc/overview/EDITING.md
dkuegler: Also just remove, I think the decision was to keep the docs of these "special edits" a bit shorter.
```suggestion
```

doc/overview/EDITING.md
dkuegler: Also, I am not really sure how these labels would propagate back into aseg.mgz and aparc.DKTatlas+aseg.mgz

m-reuter: They would not, they are only used to increase intensity in thin WM regions to avoid loosing a gyrus or parts of a lobe (usually temporal lobe) due to low grey white contrast there. Let's discuss.

recon_surf/recon-surf.sh
```diff
@@ -87,6 +87,12 @@ FLAGS:
                             that segmentations produced otherwise are conformed.
                             Requires an ABSOLUTE Path! Default location:
                             \$SUBJECTS_DIR/\$sid/mri/aparc.DKTatlas+aseg.deep.mgz
+  --mask_name <mask_file> Path to the brainmask file to use. Default location:
```
m-reuter:
```suggestion
  --mask_name <mask_file> Path to the brain mask file to use. Default location:
```

PR #593, opened 2024-10-11
Feature/tal reg long edits

neginshirvani:
This is a rebase on #592 PR and should be merged to that.

Prepare and clean up Talairach registration for better longitudinal process.

### DIFF
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -548,44 +548,22 @@ fi
 
 # ============================= TALAIRACH ==============================================
 
-if [ "$long" == "1" ] ; then
-  #TODO: move this processing into talairach-reg.sh for consistency.
+if [[ ! -f "$mdir/transforms/talairach.lta" ]] || [[ ! -f "$mdir/transforms/talairach_with_skull.lta" ]] ; then
+  # if talairach registration is missing, compute it here
+  # this also creates talairach.auto.xfm and talairach.xfm and talairach.xfm.lta
+  # all transforms (also ltas) are the same
+  cmda=("$binpath/talairach-reg.sh" "$LF"
+        --dir "$mdir" --conformed_name "$mdir/orig.mgz" --norm_name "$mdir/orig_nu.mgz")
+  if [[ "$long" == "1" ]] ; then cmda+=(--long "$basedir") ; fi
+  if [[ "$atlas3T" == "true" ]] ; then cmda+=(--3T) ; fi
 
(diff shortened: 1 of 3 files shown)


recon_surf/recon-surf.sh
```diff
@@ -556,44 +548,22 @@ fi
 
 # ============================= TALAIRACH ==============================================
 
-if [ "$long" == "1" ] ; then
-  #TODO: move this processing into talairach-reg.sh for consistency.
```
m-reuter, line 552: Why was this necessary? For long we do not run any talairach, so instead we copy things. this should not be part of a talairach-reg script as it complicates things and makes debugging more difficult.  Happy to be convinced otherwise, but I see no good reason for this.

m-reuter:
I merged the previous PR into dev, so this can be rebased to better see the additional changes that it proposes.

recon_surf/recon-surf.sh
dkuegler: My logic was that talairach registration can be called from either segmentation or surface and this way it always gets created in the right way, i.e. if people specify `--tal_reg`, it would before do the full talairach registration here: https://github.com/Deep-MI/FastSurfer/blob/4de8b3af5866f891561b6cef4ae057412dc968e3/run_fastsurfer.sh#L833

So my logic was that if the talairach registration script was clear, and would do the right thing for longitudinal (i.e. copy and have a `--long` flag) that would actually be easier to maintain and do... talairach registration stuff is always in that script and that script knows how to do it for longitudinal.

If you disagree, maybe we should change this back.

m-reuter: I see. I people want to do only segmentation, then they (at least currently) will not do any longitudinal processing, but simply run cases separately. The only reasons for switching on longitudinal in seg_only are: 
- get segmentations in same voxel space (the midspace of the base), this is usually not needed
- get a fixed eTIV across time, instead they could also just average, or median the eTIVs and fix it that way. 
Given that there is no other advantage at the moment, there is no need for a long stream in seg_only. 

But since we may move talairach to the seg_only block by default in the future, when it is faster, we can keep it like this for now.

dkuegler: The alternative would have been to add an extra check in run_fastsurfer, that ALWAYS disables --tal_reg for --long processing. This would also have addressed this specific problem.
My idea was to put things concerning talairach registration into that script and solving it in there. (Basically just moving this whole if-else-clause into the script (also some of the steps like adding the talreg into nu.mgz were duplicated).

PR #610, opened 2024-11-20
Bump versions to those in docker container

dkuegler:
After publishing the 2.3.3 containers, requirements files have not been updated in dev.

This PR pulls these updates from stable into dev.


dkuegler:
This PR should be on stable.
I am not really sure how this can be a 1-commit PR on dev ?!
It should also be merged into dev, but I am unsure if the same commit can work for both...

PR #619, opened 2024-12-12
Edit and establish the workflow "quicktest" on github actions servers 

dkuegler:
Edit and establish the workflow "quicktest" on github actions servers (compatible with github actions)

The quicktest action is triggered manually (dispatch in actions) or directly from a PR. The former can for example be used to generate the reference data. The latter may be used to add on-demand checks.

The actual test scripts have not been modified and are not expected to seamlessly work, but processing works.

On-demand checks in the pull request are triggered by adding or removing a "quicktest" label. There is a further check to ensure the workflow is only performed if people listed as collaborators changed the label.

This PR adds/changes the github workflow files and adds actions where appropriate.
Github actions are organized in .github/actions
- build-docker
- load-docker
- load-processed
- run-fastsurfer
- run-tests

The core workflow is .github/workflows/quicktest.yml. It requires 5 secrets to be set up in the repository which include the FreeSurfer license and paths to two sample images as well as their reference processing.
The Workflow defines several jobs to enable parallel processing as well as circumvent resource limitations of github actions.

The PR also adds a couple of files to .gitignore for cleaner git messages

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -1,0 +1,67 @@
+# This workflow file is a reusable workflow that will create a docker container from the currently
+# selected branch and crates a docker image in an artifact (fastsurfer-<github-sha>-docker)
+# The build artifact includes the image in docker-image.tar and the name of the image in image_name.env
+
+name: 'Build a FastSurfer docker image'
+description: 'Build a FastSurfer image'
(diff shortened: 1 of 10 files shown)


dkuegler:
Changes to the test scripts are already started in #590 .

PR #641, opened 2025-01-31
Extension: Enable seg=N and surf=N in long_fastsurfer 

dkuegler:
Make long_fastsurfer.sh support `--threads seg=X` and `--threads surf=X` notation bx fixing/adding support to `long_prepare_template.sh`

Add messages to parallel longitudinal processing to make waiting more clear Fix some variables in `long_fastsurfer.sh`

These flags were introduced in #640 

This is tested and works as expected.

### DIFF
--- a/brun_fastsurfer.sh
+++ b/brun_fastsurfer.sh
@@ -445,8 +445,12 @@ function run_single()
   then
     cmd=("${run_fastsurfer[@]}" --t1 "$image_path" "${POSITIONAL_FASTSURFER[@]}" "${args[@]}")
     if [[ "$debug" == "true" ]] ; then echo "DEBUG:" "${cmd[@]}" ; fi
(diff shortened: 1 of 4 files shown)


dkuegler:
This still currently has two (minor) issues:
1. Longitudinal surfaces are not in-fact parallel but sequential.
2. During longitudinal surfaces (in parallel mode) all run_fastsurfer log lines are doubled

dkuegler:
Well, turns out both of these seem wrong. In fact, there is only a bug in `brun_fastsurfer.sh` that does not prefix the subject id correctly making it seem like double-logging of messages from one case. In fact, it is Single-Lifting two parallel processes :face-palm:

So in reality, I just need to fix this logging issue. 

Independent of that, I think I should probably take a look at the log files as well and see if they can be deciphered as is or if the logging of parallel processes makes them impossible to read...

dkuegler:
All issues are fixed

PR #644, opened 2025-02-04
Unify and change of segmentation/surface syntax

dkuegler:
Previous commits and versions have often used a --flag (seg|surf)=value syntax. This Syntax is replaced by --flag_(seg|surf) value. This specifically affects brun_fastsurfer.sh and srun_fastsurfer.sh, which had these Syntaxes established.

This PR still needs extensive testing.

Also, not all doc entries are updated (specifically `doc/scripts/BATCH.md` and `doc/scripts/SLURM.md`).

### DIFF
--- a/brun_fastsurfer.sh
+++ b/brun_fastsurfer.sh
@@ -67,15 +67,14 @@ iii. a list of subjects directly passed (this does not support subject-specific 
   Note, brun_fastsurfer.sh will also automatically detect being run in a SLURM JOBARRAY and split
   according to \$SLURM_ARRAY_TASK_ID and \$SLURM_ARRAY_TASK_COUNT (unless values are specifically
   assigned with the --batch argument).
---parallel_subjects [surf=|seg=][<n>]: parallel execution multiple subjects.
-  * --parallel_subjects [<n>]: a total of n parallel processes (or num_subjects, if <n> is not
-    provided). The default is serial execution, or '--parallel_subjects 1'.
-  * --parallel_subjects seg=[<m>]: activate parallel execution of segmentation.
-  * --parallel_subjects surf=[<n>]: activate parallel execution of surface reconstruction.
-    Note, to avoid out-of-memory errors when parallelizing the segmentation using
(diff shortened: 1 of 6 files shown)


brun_fastsurfer.sh
```diff
@@ -67,24 +67,25 @@ iii. a list of subjects directly passed (this does not support subject-specific
...
---parallel_subjects [surf=|seg=][<n>]: parallel execution multiple subjects.
-  * --parallel_subjects [<n>]: a total of n parallel processes (or num_subjects, if <n> is not
-    provided). The default is serial execution, or '--parallel_subjects 1'.
-  * --parallel_subjects seg=[<m>]: activate parallel execution of segmentation.
-  * --parallel_subjects surf=[<n>]: activate parallel execution of surface reconstruction.
-    Note, to avoid out-of-memory errors when parallelizing the segmentation using
-    --parallel_subjects on gpus, calculate the "safe" number of concurrent segmentations and
-    specify "<m>" with the flag '--parallel_subjects seg=<m>' in addition to
-    '--parallel_subjects surf=<n>' (together max. m+n processes will run, n/m may be "max").
+--parallel_subjects <n>|max: parallel execution of run_fastsurfer for <n> images. Creates <n>
```
m-reuter, line 70: can we just call this parallel and parallel_seg ... ? Or is that conflicting with the parallel flag from the surf run (for hemispheres)?

dkuegler: It is conflicting with the `--parallel` flag, which also is a valid flag here.

dkuegler: One other option could be `--num_processes` but I am not sure this would not get confused with `--threads` and the like.
Unfortunately, `brun_fastsurfer.sh` really has 3+ different parallel processing options that have interaction effects:
`--threads`, `--parallel`, and `--parallel_subjects*`.

dkuegler:
Fixed the documentation.

dkuegler:
Tested and fixed srun_fastsurfer and brun_fastsurfer
Fixes are only debug outputs

PR #651, opened 2025-02-06
Update the way parallelization is configured

dkuegler:
- Update the `--parallel_subjects*` flags and rename them to `--parallel*` (`brun_fastsurfer.sh`, `long_fastsurfer.sh`)
- Update the "old" `--parallel` flag (making it legacy, removing from documentation everywhere incl. all examples, but also from the "functionality of the scripts")
- Update the way `--threads` works for surface reconstruction. Before: `run_fastsurfer.sh` fixes `--threads` value, if `--parallel` is set; Now: `run_fastsurfer.sh` does nothing to the threads value for surface reconstruction, but recon-surf.sh has 2 values for threads: for joint parts and for parallel hemisphere parts; if threads is >= 2 parallel hemispheres is automatically activated and the number of threads available to each parallel hemisphere is halved.

- Update the documentation throughout.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -27,8 +27,7 @@ docker run --gpus all -v /home/user/my_mri_data:/data \
                       --rm --user $(id -u):$(id -g) deepmi/fastsurfer:latest \
                       --fs_license /fs_license/license.txt \
                       --t1 /data/subjectX/t1-weighted.nii.gz \
-                      --sid subjectX --sd /output \
-                      --parallel
+                      --sid subjectX --sd /output
 ```
 
 #### Docker Flags
@@ -46,7 +45,6 @@ docker run --gpus all -v /home/user/my_mri_data:/data \
 * The `--t1` points to the t1-weighted MRI image to analyse (full path, with mounted name inside docker: /home/user/my_mri_data => /data)
 * The `--sid` is the subject ID name (output folder name)
 * The `--sd` points to the output directory (its mounted name inside docker: /home/user/my_fastsurfer_analysis => /output)
-* The `--parallel` activates processing left and right hemisphere in parallel
 
 Note, that the paths following `--fs_license`, `--t1`, and `--sd` are __inside__ the container, not global paths on your system, so they should point to the places where you mapped these paths above with the `-v` arguments. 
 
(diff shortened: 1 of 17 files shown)


Docker/README.md
```diff
@@ -27,8 +27,7 @@ docker run --gpus all -v /home/user/my_mri_data:/data \
                       --rm --user $(id -u):$(id -g) deepmi/fastsurfer:latest \
                       --fs_license /fs_license/license.txt \
                       --t1 /data/subjectX/t1-weighted.nii.gz \
-                      --sid subjectX --sd /output \
-                      --parallel
+                      --sid subjectX --sd /output
```
m-reuter: Don't we have to put threads to 2 in order to have parallel behaviour?

m-reuter: The same further below.

dkuegler: Definitely, I mean, we can also put 2 threads as the default ....

I just removed the --parallel flags because those are not really wanted/needed any more.

doc/scripts/BATCH.md
```diff
@@ -34,49 +34,48 @@ Parallelization with `brun_fastsurfer.sh`
...
-segmentation of case 2 is started. This setting is the default and represents the manual flags `--parallel_subjects 1`.
+segmentation of case 2 is started. This setting is the default and represents the manual flags `--parallel 1`.
 
 ### Single parallel pipeline
 This mode is ideal for CPU-based processing for segmentation. It will process segmentations and surfaces in series 
 in the same process, but multiple cases are processed at the same time.
 
 ```bash
-$FASTSURFER_HOME/brun_fastsurfer.sh --parallel_subjects 4 --threads 2 --parallel
+$FASTSURFER_HOME/brun_fastsurfer.sh --parallel 4 --threads 2
```
m-reuter, line 53: for same behaviour as before threads would need to be 4?

dkuegler: Yes, but the text below is updated to reflect the behavior as "requested".

run_fastsurfer.sh
```diff
@@ -535,6 +537,19 @@ check_allow_root
 
 # CHECKS
 
+if [[ "$legacy_parallel_hemi" == 1 ]] ; then
+  echo "WARNING: The --parallel flag is obsolete and will be removed in FastSurfer 3."
+  echo "  Hemispheres are now automatically processed in parallel, if threads for surface "
+  echo "  reconstruction are more than 1 (defined via --threads 2 or --threads_surf 2)!"
+  echo "IMPORTANT NOTE: The threads behavior has also changed, --parallel used to define the"
```
```suggestion
  echo "IMPORTANT NOTE: The threads behavior has also changed, --threads used to define the"
```

PR #652, opened 2025-02-06
Small but important pre-release fixes

dkuegler:
In #649, the output format of `conform.py` is changed. This leads to a bug in recon-surf, where this output is used to read the voxel size. Using the logger meant that the file and line are prepended, which did not used to be prepended. That changes the line and the extraction command became invalid.

Now fixed by using grep in `recon-surf` and adding `setup_logging` in `conform.py`. Also added extra optional `--log` flag to write the log to a file and adding a make_parser function for future integration into documentation.

Make sure all bash read statements are started with IFS="" so initial white spaces are not dropped

Removed the extra warning of expected-to-exist files `scripts/deep-seg.log` and `scripts/BUILD.log`

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -16,7 +16,6 @@
 
 # IMPORTS
 import argparse
-import logging
 import sys
 from collections.abc import Iterable
 from enum import Enum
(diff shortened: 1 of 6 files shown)


dkuegler:
This PR is based on #651 and includes its changes. #651 should be merged first.

PR #656, opened 2025-02-11
Fixes additional flags in srun_fastsurfer and T2 processing

### LINKED ISSUE #654: srun_fastsurfer does not support --t2 flag
## Description

Currently, srun_fastsurfer does not properly support additional flags in the command. In the past, this was only tested with subject_list files like

<subject id>=<t1 file> --3T --tal_reg

But additional --t2 flag combinations with T2 images have never been tested and do not work (even above --3T example is buggy).

dkuegler:
Additional flags (including --t2) did not work properly from the subject_list file in srun_fastsurfer.sh 
The Bug is actually in stools.sh:translate_cases, where the lines are not correct parsed. 
This PR fixes this issue and properly parses the file. Also, the check to find non-existent files in stools.sh:check_subject_images.

This PR also adds a check if files are symlinks, which may cause problems with folders mounted in singularity (this needed a small change in srun_fastsurfer.sh - an additional argument to check_subject_images).

This PR fixes #654

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -623,7 +623,6 @@ def conform(
     mdc_affine = mdc_affine / np.linalg.norm(mdc_affine, axis=1)
     h1["Mdc"] = np.linalg.inv(mdc_affine)
 
-    print(h1.get_zooms())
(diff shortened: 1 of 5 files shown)


dkuegler:
Basic testing complete, now testing on a bigger dataset.

m-reuter:
looks generally ok. As long as the tests work and normal processing is not breaking, we can merge it.

PR #657, opened 2025-02-11
long processing: run diff command in if statement to catch error

m-reuter:
If images have only a very, very <0.00001 mm difference in resolution in `long_fastsurfer.sh`, it fails without a clear and interpretable error message.

This is due to 2 issues:
1. `RunIt` `exit`s the script if the command it runs returns a non-zero return code. In consequence, long_fastsurfer terminates instead of printing the proper error message.
2. Very small differences in the voxelsize get rounded in the printed output of `mri_diff` but not in the default settings of the check giving very confusing output.

Solutions:
Do not use Run it here 
Print better error messages
Use `mri_diff --res-thresh 0.000001` to detect confusing messages

### DIFF
--- a/FastSurferCNN/version.py
+++ b/FastSurferCNN/version.py
@@ -322,9 +322,8 @@ def main(
             futures["checkpoints"] = pool.submit(calculate_md5_for_checkpoints)
 
         if "+pip" in sections and not prefer_cache:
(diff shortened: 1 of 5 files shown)


dkuegler:
I have improved upon @m-reuter s code. Still needs to be pushed

Issue #653, opened 2025-02-09
Manually downloading pre-trained model file (.pth) for FastSyrfer


artemmikheev:
I am writing a C++\LibTorch based software.  Can I manually download pre-trained FastSurfer model files (pth)?
I need those files to run inference.
Thanks

m-reuter:
Sure, as long as you follow the LICENSE (and cite our work in your publications) you can of course download and use FastSurfer with the available model files. One way to do this is checkout this repo and get the checkpoint files via https://deep-mi.org/FastSurfer/dev/scripts/fastsurfercnn.download_checkpoints.html

artemmikheev:
Thank you. This is potentially for a research project at NYU. Of course, you will be cited.
So there is no a simple URL on Google Drive for model pth files?
I did not plan to use your code at all. I planned to convert pth file to torchscript, so I can load it in LibTorch-based
module.

m-reuter:
We store weights in pkl files and config in yaml. See the code and includes of download_checkpoints for paths. Or just run it to get the files.

artemmikheev:
Thank you. I see the pkl files.

What is the difference between the:
aparc_cnn_axial_v1.0.0.pkl,  aparc_cnn_axial_v2.0.0.pkl, aparc_vinn_axial_v2.0.0.pkl ?

Where can I find about the expected input tensor shape and preprocessing steps.?

Background: I have this GUI Windows-only application developed at NYU, about 3,000 users. 
  https://link.springer.com/article/10.1007/s10278-025-01404-x#Sec2
It has no installation pre-requisites; you basically copy a folder to install.
It has a brain atlas registration to MNI152 and CerebrA. So I am trying to integrate FastSurfer
and make it very easily accessible to people who don't want to deal with the Python
environment.

dkuegler:
I guess doing that is an option. I would always recommend to use a docker environment to avoid having to deal with installation details (if you have security concerns, try singularity or apptainer).
I am not really sure what and how you want to use fastsurfer, but surfaces will require compiled freesurfer binaries and we will not be able to provide the kind of support you are asking for right now. You will have port quite a bit yourself.

Some files you would need to port all in FastsurferCNN:
conform.py
run_prediction.py
inference.py
Folder models

The difference between the checkpoints is manifold. 2.0.0 checkpoints are fastsurfernVINN paper (check the innovations there). 1.0.0 is fastsurfer 1 (the fastsurfer paper).

Good luck!

[...]

PR #658, opened 2025-02-12
Fix allow_root in run_fastsurfer and long_fastsurfer

dkuegler:
The flag `--allow_root` no longer skipped the error message printed.

This PR fixes this issue, also a Warning is still printed if `--allow_root` is passed.

The reason is that BASH_ARGV is only available in debug mode (breaking the behavior in check_allow_root).

I am currently building a test docker image to test if the behavior is fixed.

This bug was reported in #653 .

### DIFF
--- a/recon_surf/functions.sh
+++ b/recon_surf/functions.sh
@@ -146,8 +146,10 @@ function check_allow_root()
     echo "  running FastSurfer as root, because it will lead to files and folders created as root."
     echo "  If you are running FastSurfer in a docker container, you can specify the user"
     echo "  with '-u \$(id -u):\$(id -g)' (see https://docs.docker.com/engine/reference/run/#user)."
-    echo "  If you want to force running as root, you may pass --allow_root to $(basename "$BASH_ARGV0")."
-    if [[ "$allow_root" != "true" ]] ; then exit 1 ; fi
+    if [[ "$allow_root" != "true" ]]; then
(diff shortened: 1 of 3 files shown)


recon_surf/functions.sh
```diff
@@ -122,23 +122,32 @@ function RunBatchJobs()
...
-    echo "ERROR: You are trying to run '$BASH_ARGV0' as root. We advice to avoid running FastSurfer"
-    echo "  as root, because it will lead to files and folders created as root."
+    if [[ "$allow_root" == "true" ]] ; then LABEL="WARNING" ; else LABEL="ERROR" ; fi
+    echo "$LABEL: You are trying to run '$(basename "$BASH_ARGV0")' as root. We recommend to avoid"
+    echo "  running FastSurfer as root, because it will lead to files and folders created as root."
     echo "  If you are running FastSurfer in a docker container, you can specify the user"
     echo "  with '-u \$(id -u):\$(id -g)' (see https://docs.docker.com/engine/reference/run/#user)."
-    echo "  If you want to force running as root, you may pass --allow_root to run_fastsurfer.sh."
-    exit 1
+    echo "  If you want to force running as root, you may pass --allow_root to $(basename "$BASH_ARGV0")."
```
m-reuter: I would put this into the if block, as it does not make sense to inform about --allow_root if the user already passes it.

dkuegler: Changing it :)

dkuegler:
@m-reuter can be merged now.

PR #659, opened 2025-02-12
Optimization for long_fastsurfer, one view only for skull stripping

dkuegler:
This PR optimizes fastsurfer so that the skull stripping needed for longitudinal registration only uses one fastsurfer view.

Currently, run_prediction cannot be configured to only run on one view. But if it could, long_prepare_template would likely only need one view to compute the brainmask. Maybe we can even deactivate aseg and asegdkt output to save up time and disk space. This would then speed up this initial step.

Step 1: Modify run_prediction so the user is able to "deselect" views that should not be included. Syntax (example) `--cfg_ax none`

Step 2 (optional) Modify run_predicrion so the user is able to "deselect" certain outputs. Syntax: no `--aseg_name` means no aseg, unsure about `--asegdkt_segfile` maybe none.

Step 3: Test how reliable different view predict the brainmask.

Step 4: implement the best view for `long_prepare_template.sh`.

Step 5: Test impact on longitudinal pipeline.

### DIFF
--- a/FastSurferCNN/run_prediction.py
+++ b/FastSurferCNN/run_prediction.py
@@ -262,7 +262,11 @@ class RunModelOnData:
             "sagittal": {"cfg": cfg_sag, "ckpt": ckpt_sag},
             "axial": {"cfg": cfg_ax, "ckpt": ckpt_ax},
         }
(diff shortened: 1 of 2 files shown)


dkuegler:
Step 1 accomplished

PR #667, opened 2025-02-28
Prevent division by zero

santiestrada32:
Prevent division by zero, when computing the std on the segstats file


dkuegler (approved):
Looks good to me!

The failed check is completely unrelated.

Issue #668, opened 2025-03-03
FastSurfer on Apple M4 does not run


gregordoerl:
Hello,
Thanks for the nice tool and good documentation! Unfortunately, I was unable to get FastSurfer to run on my Mac. Here are some details:

MacBook Pro, M4 Pro Chip, 24GB mem, macOS 15.3.1
Native installation ran fine, I created a Conda environment with python3.10 without issues.

When trying to run a subject (here, a longitudinal analysis with two timepoints, one T1 each, 0.8 mm resolution), I soon got an error. My command was: 

bash ./long_fastsurfer.sh --tid s001 --t1s $dataDir/m1/s001_T1.nii $dataDir/m2/s001_T1.nii --tpids m1 m2 --sd $dataDir --py "python3.10 -s" --device mps --3T

Here a part of the error:
/AppleInternal/Library/BuildRoots/d187755d-b9a3-11ef-83e5-aabfac210453/Library/Caches/com.apple.xbs/Sources/MetalPerformanceShaders/MPSCore/Types/MPSNDArray.mm:829: failed assertion `[MPSNDArray initWithDevice:descriptor:isTextureBacked:] Error: NDArray dimension length > INT_MAX'
/Users/gregor/anaconda3/envs/FastSurfer/lib/python3.10/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '

I'll attach the log file:
[long_fastsurfer.log](https://github.com/user-attachments/files/19051564/long_fastsurfer.log)

## Environment
When running `run_fastsurfer.sh --version all` I get:
Setting ENV variable FASTSURFER_HOME to script directory /Users/gregor/FastSurfer. 
Change via environment to location of your choice if this is undesired (export FASTSURFER_HOME=/dir/to/FastSurfer)
2.4.2+7e53343

 - Installation type: native 
 - FreeSurfer Version: 7.4.1/
 - OS: macOS, MacBook Pro, M4 Pro Chip, 24GB mem, macOS 15.3.1
 - GPU: I tried the flag '--device mps' as well as without it, got the same result 

I appreciate any help, thanks in advance.

Best, gregor

dkuegler:
On first sight, this seems to be an error outside of the FastSurfer toolbox. Maybe @m-reuter has seen something like this before.

[...]

PR #672, opened 2025-03-07
Mac compatibility: Avoid PCRE grep in recon-surf, stat formatting and mapfile 

dkuegler:
Avoid using grep -P, instead use grep -E and some bash instructions.

This PR addresses #668 and was introduced by #652.

Needs testing to see if grep 2.6 / mac versions of grep on mac support the extended grep syntax.

This PR also addresses 2 more Mac compatibility issues:
- The use of `mapfile` in run_fastsurfer and brun_fastsurfer (the latter will just not be compatible for some specific flags/arguments)
- the use of `stat` which uses `-c` on Linux (GNU coreutils version) and `-f` for similar things on macOS (BSD version)

### DIFF
--- a/recon_surf/functions.sh
+++ b/recon_surf/functions.sh
@@ -1,22 +1,17 @@
 
 # set the binpath variable
-if [ -z "$FASTSURFER_HOME" ]
-then
-  binpath="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )/"
-else
-  binpath="$FASTSURFER_HOME/recon_surf/"
+if [[ -z "$FASTSURFER_HOME" ]] ; then binpath="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )/"
+else binpath="$FASTSURFER_HOME/recon_surf/"
 fi
 export binpath
 
 # fs_time command from fs60, fs72 fails in parallel mode, use local one
-# also check for failure (e.g. on mac it fails)
-timecmd="${binpath}fs_time"
-$timecmd echo testing &> /dev/null
-if [ "${PIPESTATUS[0]}" -ne 0 ] ; then
-  echo "time command failing, not using time..."
-  timecmd=""
+# also check for failure (e.g. on mac it fails, so we cannot use it there)
+if FSTIME_LOAD=0 "${binpath}fs_time" echo testing &> /dev/null ; then timecmd="${binpath}fs_time"
+else timecmd="" ; echo "INFO: Testing fs_time was not successful, not reporting per-command runtimes."
 fi
 export timecmd
+export LC_NUMERIC="en_US.UTF-8"
 
 function check_create_subjects_dir_properties()
 {
--- a/recon_surf/functions.sh
+++ b/recon_surf/functions.sh
@@ -18,6 +18,36 @@ if [ "${PIPESTATUS[0]}" -ne 0 ] ; then
 fi
 export timecmd
 
+function check_create_subjects_dir_properties()
+{
+  # 1: subjects_dir
+  if [[ -z "$1" ]]
+  then
+    echo "ERROR: No subject directory defined via --sd. This is required!"
+    exit 1
+  elif [[ ! -d "$1" ]]
+  then
+    echo "INFO: The subject directory did not exist, creating it now."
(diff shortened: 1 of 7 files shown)


m-reuter:
This works on my Mac. I am getting a print out
```A min-conformedconformed image ... ``` which looks weird. 

Also I wonder if we should use this PR to fix the other Mac issues:
- the BSD tool ```stat``` in Mac does not have the -c flag (I think it uses -f flag for format)
- the missing ```mapfile```

**Comment by dkuegler** [comment]
> This works on my Mac. I am getting a print out `A min-conformedconformed image ... ` which looks weird.
> 
> Also I wonder if we should use this PR to fix the other Mac issues:
> 
>     * the BSD tool `stat` in Mac does not have the -c flag (I think it uses -f flag for format)
> 
>     * the missing `mapfile`

Let's use this issue for that. The grep issue I was fixing here is a mac-only issue.

I will fix the `mapfile` thing very quickly (this is just an information things, so no damage, just going to skip that for old bash versions).
I will see about `stat` as well

**Comment by dkuegler** [comment]
Still working on stat
`mapfile` changes are untested on macos

**Comment by m-reuter** [comment]
I tested mapfile changes and it works (also tested --threads 4 for parallel). 

One other thing, I get the run time only in full hours : ```#@#%# recon-surf-run-time-hours  0.000```

Probably due to first output line: ```time command failing, not using time...```
while a time command is available on Mac, probably uses a different format. 
Update: I just saw that on Linux it also only reports the full hours. So we should probably report also minutes in both cases and we should check if the time command failing printout should be changed to a message that looks less like an error or warning, maybe "unexpected time command (maybe on MAC), using ??? instead"

File: recon_surf/functions.sh [hunk 1]
```diff
@@ -1,22 +1,48 @@
...
+  elif [[ ! -d "$1" ]]
+  then
+    echo "INFO: The subject directory did not exist, creating it now."
+    if [[ "$(id -u)" == 0 ]] ; then echo "WARNING: Creating as root!" ; fi
+    if ! mkdir -p "$1" ; then echo "ERROR: directory creation failed" ; exit 1; fi
+  else
+    if [[ "$(stat --version | head -n 1)" =~ GNU[[:space:]]coreutils ]] ; then # linux
+      user_group=$(stat -c "%u:%g" "$1")
+      world_access=$(stat -c "%a" "$1" | tail -c 2)
+    elif [[ "$(stat -- 2>&1 | tail -n 1)" =~ ^usage:[[:space:]]stat.*-f[[:space:]]format ]] ; then # macOS
```
dkuegler: @m-reuter this is the line where `2>&1` does not capture stderr it seems

m-reuter:
```
> stat --version
stat: illegal option -- -
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]
```

m-reuter: also the whole block gives the same error message and stops, does not go into elif block.

m-reuter: I think this way would work:
```
if stat --version &>/dev/null; then
  echo "This is the GNU version of stat."
else
  echo "This is the BSD version of stat."
fi
```

m-reuter:
I still get ```./FastSurfer/run_fastsurfer.sh: line 828: mapfile: command not found```. Running a test now.

dkuegler:
> I still get `./FastSurfer/run_fastsurfer.sh: line 828: mapfile: command not found`. Running a test now.

What do you get, if you run `bash --version` ?

m-reuter:
inside the scripts it uses the system bash, which is this:
```
> /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)
Copyright (C) 2007 Free Software Foundation, Inc.
```
in my terminal I have a newer one (5.something) via home-brew, but system bash is still old, and that is the one that gets used via the bash shebang.

dkuegler:
So the difference is that when we call `bash --version`, it is different to `./reconsurf.sh` which uses the other bash .... This is so confusing.

m-reuter:
it can be different depending on what other bash is installed on the system and in the path. On normal systems, it should be the same.

dkuegler:
I have changed the usage of `bash --version` to `${BASH_VERSINFO[0]}` which is available in bash 3.2 according to the docs of bash 3.2.

Issue #676, opened 2025-03-13
How to use this FastSurfer as the a SynthSeg from FreeSurfer to obtain brain regions volumes by segmentation


JornyWan:
Hi, FastSurfer is a great work that i can use to replace FreeSurfer, while I am looking for features that could get brain regions volumes just as **mri_synthseg** and get the volumes results (and is it possible to keep them without saving them and just use the torch tensor to keep them?)

Thanks!

m-reuter:
Hi, the easiest way is of course to call run_fastsurfer.sh with appropriate flags ( e.g. ```--seg-only ```) which will run all all our different segmentation modules (full brain, cerebellum sub segmentation and hypothalamus segmentation) sequentially. You can also switch off individual networks via flags. 

If you want to directly call routines from python, you can of course do that. Take a look at run_prediction https://github.com/Deep-MI/FastSurfer/blob/dev/FastSurferCNN/run_prediction.py which performs the segmentation for full brain. 

And for getting the volumes from the segmentation, you can simply add up the voxels and multiply by voxel size, or use our segstats tool, which also accounts for partial volume effects.

JornyWan:
thanks, i think i only need to have similar functions as synthseg has, to get the volumetric measurement, so is it like i only need to run the run_predictions right?

dkuegler:
I would recommend you run `run_fastsurfer.sh --no_cereb --no_hypothal --seg_only` and then just delete the folder. FastSurfer really is not made to be called from python in a pipeline, but what you are looking for is the function `get_prediction` in https://github.com/Deep-MI/FastSurfer/blob/412360814aa6664395ffabf3cdf1fc1f7d977bb6/FastSurferCNN/run_prediction.py#L369

You will have to do the preparation on eval (the inference object yourself, so take a look at the `main` function in `run_prediction.py`.

Also note, that the python interfaces should not considered "stable" at this point, they may (and will) change in future releases (for example to enable pip integration).

JornyWan:
thanks!!! I just run the FastSurfer with singularity for one 3D MRI, and the brain segmentation looks well, but it seems like the stats report does not have the grey matter region recorded.

m-reuter:
You find the stats in the ```<subjectd_dir>/<subject_id>/stats/aseg+DKT.stats```

In the table you find the stats for subcortical grey matter structures such as hippocampus etc, and also for all cortical regions "ctx-lh-...". 

[...]

PR #678, opened 2025-03-14
Add additional stats/aseg.VINN.stats file

dkuegler:
This PR adds a new stats file for the segmentation-based `mri/aseg.no_CC.mgz` in `stats/aseg.VINN.mgz`.
This includes the Left/Right-Cerebral-Cortex labels in the table. -- The TotalGray volume (avaliable after surfaces) is still not available (this is in the header with Measures), but depends on the Cortical Gray Matter volumes (which depends on the surface-derived volume). Because of differing conventions on how this is computed, we do not mix this here.

Also renames the statsfile `stats/aseg+DKT.mgz` computed from `mri/aparc.DKTAtlas+aseg.deep.mgz` to `stats/aseg+DKT.VINN.mgz`. `stats/aseg+DKT.mgz` is generated as a softlink for compatibility (pointing to `stats/aseg+DKT.VINN.stats`).

---
## The previous approach is described below.
The problem is that these values cannot be loaded with asegstats2table.

This PR adds "sums" over Cortical GM in Measures as talked about in #676.

- Add CortextVol, lhCortexVol, rhCortexVol, TotalGrayVol as keys for measures
- Add these measures into run_fastsurfer for aseg+DKT.stats

Example of content update of `aseg+DKT.stats`.

```
# Measure Mask, MaskVol, Mask Volume, 1747144.000000, mm^3
# Measure BrainSeg, BrainSegVol, Brain Segmentation Volume, 1385005.314630, mm^3
# Measure BrainSegNotVent, BrainSegVolNotVent, Brain Segmentation Volume Without Ventricles, 1313802.684199, mm^3
# Measure SupraTentorial, SupraTentorialVol, Supratentorial volume, 1228319.463260, mm^3
# Measure SupraTentorialNotVent, SupraTentorialVolNotVent, Supratentorial volume, 1157116.832829, mm^3
# Measure SubCortGray, SubCortGrayVol, Subcortical gray matter volume, 66030.583679, mm^3
# Measure rhCortexVol, rhCortexVol, Right hemisphere cortical gray matter volume (segmentation-based), 271607.121787, mm^3
# Measure lhCortexVol, lhCortexVol, Left hemisphere cortical gray matter volume (segmentation-based), 265131.161238, mm^3
# Measure CortexVol, CortexVol, Total cortical gray matter volume (segmentation-based), 536738.283025, mm^3
# Measure TotalGrayVol, TotalGrayVol, Total gray matter volume (all segmentation-based), 728047.728233, mm^3
# Measure rhCerebralWhiteMatter, rhCerebralWhiteMatterVol, Right hemisphere cerebral white matter volume, 277147.639639, mm^3
# Measure lhCerebralWhiteMatter, lhCerebralWhiteMatterVol, Left hemisphere cerebral white matter volume, 277477.805146, mm^3
# Measure CerebralWhiteMatter, CerebralWhiteMatterVol, Total cerebral white matter volume, 554625.444785, mm^3
```

### DIFF
--- a/run_fastsurfer.sh
+++ b/run_fastsurfer.sh
@@ -451,7 +451,8 @@ case $key in
     fi
     run_asegdkt_module="0"
     ;;
-  --asegdkt_statsfile) asegdkt_statsfile="$1" ; shift ;;
+  --asegdkt_statsfile) asegdkt_vinn_statsfile="$1" ; shift ;;
+  --aseg_statsfile) aseg_vinn_statsfile="$1" ; shift ;;
   --aseg_segfile) aseg_segfile="$1" ; shift ;;
   --mask_name) mask_name="$1" ; warn_seg_only+=("$key" "$1") ; warn_base+=("$key" "$1") ; shift ;;
   --merged_segfile) merged_segfile="$1" ; shift ;;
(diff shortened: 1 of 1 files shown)


dkuegler:
I looked into asegstats2table and the format I have chosen here is not compatible with that script.

There is also no easy way to resolve this ambiguity, three options:
1. use the `lhCortex` Measure that is usually filled from the surfaces (so it would not be clear that this is volume-based)
2. put a Cortical area line into the table (this would mean the "sum" of all regions might be high, but this property did not hold before either anyways, because the cortex was not in aseg.stats before
3. Generate a second file for aseg.stats -- maybe aseg.deep.stats ?

dkuegler:
Ready for review. Current state is tested and runs with expected outputs.

run_fastsurfer.sh
```diff
@@ -965,24 +968,19 @@ then
...
-                 2027 2028 2029 2030 2031 2034 2035
-           --lut "$fastsurfercnndir/config/FreeSurferColorLUT.txt"
-           measures --compute "Mask($mask_name)" "BrainSeg" "BrainSegNotVent"
-                              "SupraTentorial" "SupraTentorialNotVent"
-                              "SubCortGray" "rhCerebralWhiteMatter"
-                              "lhCerebralWhiteMatter" "CerebralWhiteMatter"
-           # make sure to read white matter hypointensities from the
-           )
+           --ids 2 4 5 7 8 10 11 12 13 14 15 16 17 18 24 26 28 31 41 43 44 46 47 49 50 51 52 53 54 58 60 63 77
+                 251 252 253 254 255
```
m-reuter, line 975: These are CC ids, I thought they are not in the VINN segmentation and therefore we can omit them for now.

dkuegler: Yes, this is the "aseg+DKT.stats" file. It is replicating the aseg format and has therefore had CC ids in there since the beginning.
I can obviously remove them, but why change the format of an existing and established file.

PR #684, opened 2025-03-21
Update doc for Docker in the rootless mode

plu-project:
This pull request adds solution for users using Docker in the rootless mode. For users using Docker in the rootless mode, the existing Docker installation and the example command prevents FastSurfer from running.

Original Problems:
1. There is an additional need to install the NVIDIA container toolkit and configure it for Docker in the rootless mode. Otherwise, this error message raises: "docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]]".
2. In the example command, the --user flag mapping with the group- and user-id leads to the user in the container as "I have no name!" and therefore the user in the container does not have the write-access to the mounted output directory, even though the output directory in the host system has given permission to everyone. 

To be noted:
The "--user 0" in the example command in the pull request is needed for Docker in the rootless mode. Without this, Docker complains "Error response from daemon:unable to find user nonroot".


doc/overview/EXAMPLES.md
```diff
@@ -25,7 +25,7 @@ Docker Flags:
...
 * The `--rm` flag takes care of removing the container once the analysis finished. 
-* The `--user $(id -u):$(id -g)` part automatically runs the container with your group- (id -g) and user-id (id -u). All generated files will then belong to the specified user. Without the flag, the docker container will be run as root which is discouraged.
+* The `--user $(id -u):$(id -g)` part automatically runs the container with your group- (id -g) and user-id (id -u). All generated files will then belong to the specified user. Without the flag, the docker container will be run as root which is discouraged. However, if you use Docker in the **rootless mode** with your group- and user-id, FastSufer does not have the write permission for the output directory. For the rootless mode, you should set `--user 0` and add the `--allow_root` flag. There will be no write permission issue for the output directory and you will be the owner of files created by FastSurfer.
```
dkuegler:
```suggestion
* The `--user $(id -u):$(id -g)` part automatically runs the container with your group- `id -g` and user-id `id -u`. All generated files will then belong to the specified user. Setting a user id is required! Running docker as root is discouraged. Note, that in the **rootless mode**, the operating system implements the translation to *your* user- and group id. Therefore, for rootless mode,  you must set `--user 0` and add the FastSurfer `--allow_root` flag!
```

dkuegler (approved):
Thank you for your contribution!

PR #691, opened 2025-05-08
Add segmentation of images in "native image space"

dkuegler:
Add segmentation of images in "native image space". 

This adds the `--native_image` flag to `run_fastsurfer.sh` and multiple more fine-grained flags to `conform.py` and the various `run_predicions.py`s for more fine control over the 3 components of the image space: "orientation", "voxel size" and "image size".

`--orientation`
---
For `conform.py`, this commit implements additional options for orientation in addition to the previous "lia" which is default and required for downstream surface reconstruction with FreeSurfer binaries.
--orientation "XXX" e.g. "lia" -- reformat the affine so it is always aligned along the primary axes, for example [-X 0 0, 0 0 -X, 0 X 0] with X the voxel size for "lia",
--orientation "soft-XXX", e.g. "soft-lia" -- reformat the affine so the primary directions as above, but axes might not be perfectly aligned. For this, DOES NOT -- by itself -- trigger an interpolate the image.
-- orientation "native" -- do not reformat the affine, keep it as is.

There is a second change to native images done by conform, changing the image dimensions so all dimensions in all directions are the same (e.g. 256 for 1mm voxelsize images), but this number is different for highres images.
This behavior is still the default behavior (explicitly also --image_size "auto").
However, as this will also change the native image dimensions for segmentation, additional options are available:
--image_size "auto": as before/above
--image_size int: directly force a specific image dimension, e.g. 256.
--image_size "fov": keep the field of view consistent, i.e. if the voxel size is not touched, keep the image size the same, or -- if the voxel size is changed via --vox_size -- set the image size such that the field of view -- the region we have data for -- stays consistent.

Changes to files:
conform.py and its functions are heavily restructured to optimally support these features.
run_prediction and run_fastsurfer get --orientation and --image_size flags (run_fastsurfer also gets the --keepgeom flag: alias for --orientation native --image_size fov)
add and consistently use function to change data orientations
add documentation for new flags


FastSurferCNN/data_loader/conform.py
```diff
@@ -95,75 +93,91 @@ def make_parser() -> argparse.ArgumentParser:
...
+        type=__img_size,
+        help="Specifies the image size to conform to (deactivated by 'any'). Also allows 'fov' for conforming to the "
+             "field of view.",
+    )
+    parser.add_argument(
+        "--rescale",
+        dest="rescale",
+        default="auto",
+        choices=("auto", "robust", "none"),
+        help="Specifies whether image intensities should be rescaled, 'robust' will robustly intensities to the output "
```
m-reuter: I don't understand the default here. from the description it seems that auto is the same as robust?

m-reuter: OK, becomes clear later. It should say in the description that robust depends on the --dtype flag, while auto always overwrites --dtype to uchar? or what?

FastSurferCNN/data_loader/conform.py
```diff
@@ -95,75 +93,91 @@ def make_parser() -> argparse.ArgumentParser:
...
     )
     advanced.add_argument(
-        "--no_strict_lia",
-        dest="force_strict_lia",
-        action="store_false",
-        help="Ignore the forced LIA reorientation.",
+        "--orientation",
+        dest="orientation",
+        default="lia",
+        choices=("lia", "soft-lia", "soft lia", "native"),
```
m-reuter: not happy with having both soft-lia and soft lia. Also (we need to check) I think LIA and RAS (usually capitalised) are also used as a general slicing direction, therefore we might want to consider calling Lia -> strict-lia and soft-lia just lia ?

FastSurferCNN/data_loader/conform.py
```diff
@@ -95,75 +93,91 @@ def make_parser() -> argparse.ArgumentParser:
...
+        help="Specify the target data orientation."
     )
     advanced.add_argument(
-        "--no_lia",
-        dest="force_lia",
-        action="store_false",
-        help="Ignore the reordering of data into LIA (without interpolation). "
-             "Supersedes --no_strict_lia",
+        "--no_strict_lia", "--allow_soft_lia",
```
m-reuter: why is there a second flag? should this not be handled with --orientation ?

FastSurferCNN/data_loader/conform.py
```diff
@@ -95,75 +93,91 @@ def make_parser() -> argparse.ArgumentParser:
...
+        action="store_const",
+        const="soft lia",
+        help="Ignore the forced LIA reorientation and only reorient to a soft LIA orientation.",
     )
     advanced.add_argument(
-        "--no_iso_vox",
-        dest="force_iso_vox",
-        action="store_false",
-        help="Ignore the forced isometric voxel size (depends on --conform_min).",
+        "--no_lia", "--allow_native",
```
m-reuter: and another lia flag. I think this is very confusing.

FastSurferCNN/data_loader/conform.py
```diff
@@ -841,9 +974,9 @@ def is_lia(
     ----------
     affine : np.ndarray
         The affine to check.
-    strict : bool, default=True
-        Whether the orientation should be "exactly" LIA or just similar to LIA (i.e.
-        it is more LIA than other directions).
+    soft : bool, default=True
+        Whether the orientation are required to be "exactly" LIA or just similar (soft) (i.e. it is roughly oriented as
```
m-reuter:
```suggestion
        Whether the orientation is required to be "exactly" LIA or just similar (soft) (i.e. it is roughly oriented as
```

FastSurferCNN/utils/parser_defaults.py
```diff
@@ -298,6 +299,25 @@ class SubjectDirectoryConfig:
              "determines the voxel size from the image itself (conforming to the minimum voxel size, or 1 if the "
              "minimum voxel size is above 0.95mm). ",
     ),
+    "orientation": __arg(
+        "--orientation",
+        choices=get_args(OrientationType),
+        type=__orientation,
+        dest="orientation",
+        default="lia",
+        help="Select the target affine format for output, native: input defined by input image, soft_lia (or soft-lia, "
```
m-reuter: not happy with soft_lia ( soft-lia , soft_lia ) . we should stick to one of them I think.

doc/overview/FLAGS.md
```diff
@@ -32,6 +32,9 @@ In the following, we give an overview of the most important options. You can vie
 * `--no_cereb`: Switch off the cerebellum sub-segmentation.
 * `--cereb_segfile`: Name of the cerebellum segmentation file. If not provided, this intermediate DL-based segmentation will not be stored, but only the merged segmentation will be stored (see --main_segfile <filename>). Requires an ABSOLUTE Path! Default location: \$SUBJECTS_DIR/\$sid/mri/cerebellum.CerebNet.nii.gz
 * `--no_biasfield`: Deactivate the biasfield correction and calculation of partial volume-corrected statistics in the segmentation modules.
+* `--orientation`: **Only supported for `--seg_only`**, select the image orientation (`lia` (default), `soft lia` or `soft-lia` -- orient to LIA orientation, but do not interpolate, `native` -- segment in the native image space).
```
m-reuter: again here

doc/overview/FLAGS.md
```diff
@@ -32,6 +32,9 @@ In the following, we give an overview of the most important options. You can vie
...
+* `--image_size`: Select the target image size (a number to use, `fov` to set the image size to the input field of view (**allows non-cube images, only supported for `--seg_only`**), or `auto` (default) expands the image so all three image dimensions are the same).
+* `--keepgeom`: **Only supported for `--seg_only`**, segment in native image space (both orientation and image size), same as `--orientation native --image_size fov`.
```
m-reuter: we should consider --keep_geom

FastSurferCNN/data_loader/conform.py
dkuegler: I should word this better.

"auto" (default): depends on the `--dtype flag",  if the target dtype is uint8 (the default), => like "robust" else "none"
"robust": robustly rescale intensities to 0-maxvalue of dtype (e.g. 0-255 for uint8)
"none": disabled/no rescaling

This never overwrites the target dtype

FastSurferCNN/data_loader/conform.py
dkuegler: we can change it to just soft-lia or soft_lia... this is just there because the space can be a problem

I would not have different meanind depedning on capitalization... that sounds like a guaranteed thing for confusion.

FastSurferCNN/data_loader/conform.py
dkuegler: that is legacy, see --no_strict_lia in line 144 in the history, and yes, in the future this should just be --orientation, but for now this is fine as an advanced flag.

FastSurferCNN/data_loader/conform.py
dkuegler: see above (I can add that this is legacy to the help text)

FastSurferCNN/utils/parser_defaults.py
dkuegler: Lets discuss this. I am most happy wioth just `soft_lia` I think, but we can use any of the different versions (right now `soft_lia`, `soft lia` and `soft-lia` are all the same.

doc/overview/FLAGS.md
dkuegler: What do you mean? do we not consider `--keepgeom`?

FastSurferCNN/data_loader/conform.py
dkuegler: I have checked mri_convert and it has an option to convert the orientation:

```bash
$ mri_convert -i T1_0.mgz -o T1_lia0.mgz --out_orientation LIA
mri_convert -i T1_0.mgz -o T1_lia0.mgz --out_orientation LIA 
reading from T1_0.mgz...
TR=2560.00, TE=0.00, TI=0.00, flip angle=0.00
i_ras = (-0.997722, 0.0353588, 0.0574437)
j_ras = (0.0257398, 0.986732, -0.160305)
k_ras = (0.0623497, 0.158461, 0.985395)
Setting output orientation to LIA
Reslicing using trilinear interpolation 
writing to T1_lia0.mgz...
$ mri_info T1_lia0.mgz 
[...]
voxel to ras transform:
               -0.8000   0.0000  -0.0000    93.1617
                0.0000   0.0000   0.8000  -114.3121
                0.0000  -0.8000  -0.0000   123.1322
                0.0000   0.0000   0.0000     1.0000
```

So strict-lia is less consistent with FreeSurfer, because mri_convert --out_orientation lia (also) leads to strict lia. That is why I have also changed the terminology etc to "soft" over "strict" in most places.

FastSurferCNN/data_loader/conform.py
dkuegler: If you want to remove these legacy options now, that can be done. They were introduced in 3fdf6ffb10e05dc1850823f59dddf64cf02a5590, i.e. with 2.3

dkuegler:
I think this should fix everything, but I will run a final check for non-isotropic images tomorrow.

FastSurferCNN/run_prediction.py
```diff
@@ -325,28 +336,23 @@ def conform_and_save_orig(
...
-                conform_vox_size=self.vox_size,
-                conform_to_1mm_threshold=self.conform_to_1mm_threshold,
-            )
+            self.async_save_img(subject.copy_orig_name, orig_data, orig, orig_data.dtype)
+
+        if not is_conform(orig, **self.__conform_kwargs(verbose=True)):
+            if (self.orientation is None or self.orientation == "native") and \
+                    not is_conform(orig, **self.__conform_kwargs(verbose=False, dtype=None)):
+                raise RuntimeError(
+                    f"To store images in native image space, the input image {subject.orig_name} must have isometric "
```
m-reuter, line 345:
```suggestion
                    f"To store images in native image space, the input image {subject.orig_name} must have isotropic "
```

doc/overview/FLAGS.md
m-reuter: I think in documentation we should use native_image as it is easier to understand.

run_fastsurfer.sh
```diff
@@ -695,6 +711,12 @@ then
   fi
 fi
 
+if [[ "$run_surf_pipeline" == "1" ]] && [[ "$native_image" != "false" ]]
+then
+  echo "ERROR: The surface pipeline is not compatible with the options --native_image and "
```
```suggestion
  echo "ERROR: The surface pipeline is not compatible with the options --native_image or "
```

FastSurferCNN/data_loader/conform.py
```diff
@@ -85,99 +113,119 @@ def make_parser() -> argparse.ArgumentParser:
...
+    )
+    parser.add_argument(
+        "--img_size",
+        dest="img_size",
+        default="auto",
+        metavar="<int>|auto|fov|any",
+        type=__img_size,
+        help="Specifies the image size to conform to. Options: <int> (dimension of the target image), 'auto' "
+             "(determine dimensions of image from largest field-of-view dimension, min. 256), 'fov' (determine "
+             "dims. of image from field of view, dims. may be differ) or 'any' (ignore this criteria, similar to fov).",
```
m-reuter: not toally clear here what the difference between auto and fov is. maybe auto is always a cube ?

FastSurferCNN/run_prediction.py
m-reuter: also check if you find other occurrences of "isometric".

doc/overview/FLAGS.md
dkuegler: I actually forgot to change this here. `--orientation` and `--image_size` no longer supported, just `--keepgeom` and `--native_image`.

FastSurferCNN/run_prediction.py
dkuegler: I will change that, but probably in a different PR. I have one lined up with only formatting changes. I am actually only testing that one as it is rebased on this PR.

FastSurferCNN/data_loader/conform.py
dkuegler: "auto" is basically fov but expanded to a cube, "fov" keeps the original field of view

PR #697, opened 2025-05-21
Fixes so de-selecting views for processing for run_prediction is possible

dkuegler:
Modify run_prediction so the user is able to "deselect" views to not include.

Details:
- parser_defaults.py: Allow empty or none for --cfg_ax, --cfg_sag, --cfg_cor.
- run_prediction.py: If cgf_ax/sag/cor is none/empty, do not use that view

This is also Step 1 for #659, but it is also usefull in some anisotropic voxelsizes settings.


dkuegler:
@m-reuter bugs are fixed, works now.

PR #700, opened 2025-05-27
Bugfixes before release

dkuegler:
This PR fixes several bugs that came up and were resolved during release testing and further review.

Fix the usage of `nibabel.orientations.ornt_transform` after clarification. https://github.com/nipy/nibabel/issues/1418

Fix Argument-Type related bugs.
- "soft LIA" to "soft-LIA"
- ArgumentTypeError to ValueError
- vox_size can accept a float in addition to a str
- fix image_size default value overwriting passed value
- remove test from Docker
- reorder vox_size and image_size check in is_conform so their order in output is swapped
- set LANG to en_US in Dockerfile
- fix image position when cropping with crop_transform
- only reset LC_NUMERIC in recon_surf/functions, if LC_NUMERIC is not en_US

### DIFF
--- a/FastSurferCNN/utils/arg_types.py
+++ b/FastSurferCNN/utils/arg_types.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
 from itertools import permutations, product
 from typing import Literal, cast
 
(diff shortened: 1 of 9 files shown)


FastSurferCNN/utils/arg_types.py
```diff
@@ -201,13 +202,13 @@ def target_dtype(a: str) -> str:
     msg = "The following dtypes are verified: " + ", ".join(dtypes)
     if np.dtype(_a).name == _a:
         # numpy recognizes the dtype, but nibabel probably does not.
-        print(
-            f"WARNING: While numpy recognizes the dtype {a}, nibabel might not and this might lead to compatibility "
-            f"issues. {msg}"
-        )
+        from logging import getLogger
+
+        warnmsg = f"While numpy recognizes the dtype {a}, nibabel might not leading to compatibility issues. "
```
m-reuter, line 207:
```suggestion
        warnmsg = f"While numpy recognizes the dtype {a} nibabel might not, leading to compatibility issues. "
```

recon_surf/functions.sh
```diff
@@ -11,7 +11,8 @@ if FSTIME_LOAD=0 "${binpath}fs_time" echo testing &> /dev/null ; then timecmd="$
 else timecmd="" ; echo "INFO: Testing fs_time was not successful, not reporting per-command runtimes."
 fi
 export timecmd
-export LC_NUMERIC="en_US.UTF-8"
+if [[ "$LC_NUMERIC" != "en_US.UTF-8" ]] ; then LC_NUMERIC="en_US.UTF-8" ; fi
+export LC_NUMERIC
```
m-reuter: In how far is the new code different from the old? in all cases is it set to en_US and exported either way?

Issue #587, opened 2024-10-03
conforming adding header info


m-reuter:
## Description

When conforming happens from inside run_prediction it adds incorrect "TR" information to the image header. 

## Steps to Reproduce

Run run_prediction with an already conform input, e.g. from oasis and compare the conformed output image. 
mri_info will show on the original:
            `TR: 0.00 msec, TE: 0.00 msec, TI: 0.00 msec, flip angle: 0.00 degrees`
and on the conformed:
            `TR: 1000.00 msec, TE: 0.00 msec, TI: 0.00 msec, flip angle: 0.00 degrees`
So acquisition parameters change, while the image is identical (no conforming necessary as it is already). 

## Expected Behavior

Copy those values from the original header. 

## Environment
 - FastSurfer Version: dev
 - OS:  macos

absl-py==2.1.0
certifi==2024.8.30
charset-normalizer==3.3.2
click==8.1.7
contourpy==1.3.0
cycler==0.12.1
Deprecated==1.2.14
filelock==3.16.0
fonttools==4.53.1
fsspec==2024.9.0
grpcio==1.66.1
h5py==3.11.0
humanize==4.10.0
idna==3.8
imageio==2.35.1
Jinja2==3.1.4
joblib==1.4.2
kiwisolver==1.4.7
lapy==1.1.1
lazy_loader==0.4
Markdown==3.7
markdown-it-py==3.0.0
MarkupSafe==2.1.5
matplotlib==3.9.2
mdurl==0.1.2
mpmath==1.3.0
networkx==3.3
nibabel==5.2.1
numpy==1.26.4
packaging==24.1
pandas==2.2.2
pillow==10.4.0
plotly==5.24.0
protobuf==5.28.0
psutil==6.0.0
Pygments==2.18.0
pyparsing==3.1.4
python-dateutil==2.9.0.post0
pytz==2024.1
PyYAML==6.0.2
requests==2.32.3
rich==13.8.0
scikit-image==0.24.0
scikit-learn==1.5.1
scipy==1.14.1
shellingham==1.5.4
SimpleITK==2.4.0
six==1.16.0
sympy==1.13.2
tenacity==9.0.0
tensorboard==2.17.1
tensorboard-data-server==0.7.2
threadpoolctl==3.5.0
tifffile==2024.8.30
torch==2.4.1
torchio==0.19.9
torchvision==0.19.1
tqdm==4.66.5
typer==0.12.5
typing_extensions==4.12.2
[...]

PR #701, opened 2025-05-30
Fix conform.py losing tr value during conversion

OtabekRintaro:
Previously, when conform.py had nifti file as input, it wouldn't copy the tr information into conformed output image.
Reason: MGHheader and niftiheader formats store this information differently.


m-reuter:
Thanks for your contribution. Maybe this is also related to #587

Do you know of any documentation about this online? E.g. why is the info in pixdim 4 and needs multiplication by 1000?

OtabekRintaro:
Sadly, I didn't find a precise definition in documentation of Nifti's specification website.
I have scraped some partial definitions from different sources, here they are:

**Nifti Specification FAQ**
 https://nifti.nimh.nih.gov/nifti-1/documentation/faq.html
 Q20. How is the timing of acquisition encoded in the NIfTI-1 header? 
 ...
 slice_duration * dim[slice_dim] can be less than pixdim[4] with a clustered acquisition method, for example.

 so there is a correlation between slice_duration (timing of acquisition) encoding and pixdim[4]
 
 **Nifti Specification**
 https://nifti.nimh.nih.gov/nifti-1/documentation/Analyze_usage.pdf
 This file which shows how different FMRI software packages use header values which contains the line that pixdim[4] = TR.
 
 **Documentation by NITRC**
 https://www.nitrc.org/docman/view.php/26/64/nifti1.h
  The codes below can be used in xyzt_units to indicate the units of pixdim.
   As noted earlier, dimensions 1,2,3 are for x,y,z; dimension 4 is for
   time (t).
    - If dim[4]=1 or dim[0] < 4, there is no time axis.
    - A single time series (no space) would be specified with
      - dim[0] = 4 (for scalar data) or dim[0] = 5 (for vector data)
      - dim[1] = dim[2] = dim[3] = 1
      - dim[4] = number of time points
      - pixdim[4] = time step
 
 **One of the chapters in course about fMRI data analysis from Berkley**
 https://bic-berkeley.github.io/psych-214-fall-2016/tr_and_headers.html
 
 You can get the voxel (plus TR) dimensions with the get_zooms method of the header object:
 > func_img = nib.load('ds114_sub009_t2r1.nii')
 > header = func_img.header
 > header.get_zooms()
 (4.0, 4.0, 4.0000162, 2.5)
 In this case, the image spatial voxel sizes are (4 by 4 by 4) millimeters, and the TR is 2.5 seconds.
 
 We can see, that the TR is specified in seconds, when mgz files store the value in ms, so we need to convert seconds into milliseconds.
 
 Also, by reading documentations, I have realized that conform.py does well while converting from nii.gz/nii to mgz but it still loses tr value due to it not having a 'tr' value and nib.save doesn't do the copy by itself. Basically, when nib.save converts mgh header to nii.gz header then it doesn't copy the tr into pixdim[4]. I will see if there is a fix for that as well.

Issue #712, opened 2025-06-20
Ruff error (optional)


m-reuter:
Ruff is currently throwing a few errors: 
https://github.com/Deep-MI/FastSurfer/actions/runs/15774829979/job/44466783376

that need fixing.

PR #713, opened 2025-06-20
Fix ruff error suppression of typing.Optional

OtabekRintaro:
Problem: 
Ruff rules are not suppressed. Specifically, typing.Optional, previously UP007.

Cause:
According to Ruff's [latest release's notes](https://github.com/astral-sh/ruff/releases/tag/0.12.0): 

> non-pep604-annotation-union (UP007) has now been split into two rules.
> UP007 now applies only to typing.Union, while
> non-pep604-annotation-optional (UP045) checks for use of
> typing.Optional. UP045 has also been stabilized in this release, but you
> may need to update existing include, ignore, or noqa settings to
> accommodate this change.

Issue: https://github.com/Deep-MI/FastSurfer/issues/712


m-reuter:
thanks, can you also fix the spell check error (just replace "softwares" by "packages")
Also instead of suppressing RUFF rules, it would be better to fix the underlying problem and use the newer convention (unless that has other consequences).

PR #760, opened 2026-01-05
Documentation better linking of symbols to files and cleanup

dkuegler:
This PR fixes an issue with the doc build, where symbols could not be found/links could not be created in the CC documentation.
As a side effect, this PR also fixes some other minor things in the documentation build, including linking to the correct branch in the stable doc, getting rid of extra dependencies in the doc optional dependencies, and adding links in other spots when possible.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,9 +1,5 @@
-.. FastSurfer documentation master file, created by
-   sphinx-quickstart on Thu Nov 30 15:48:44 2023.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
-
-   Modified and updated by the David Kügler of the FastSurfer Team.
+.. FastSurfer documentation master file; modified and updated by
+   David Kügler of the FastSurfer Team.
 
 .. image:: images/teaser.png
    :alt: FastSurfer Teaser Image
--- a/FastSurferCNN/version.py
+++ b/FastSurferCNN/version.py
@@ -1,6 +1,7 @@
 #!/bin/python
 
 import argparse
+import io
 import re
 import shutil
 import subprocess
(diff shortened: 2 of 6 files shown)


.github/workflows/quicktest.yaml
```diff
@@ -166,7 +166,7 @@ jobs:
             fi
             echo "FASTSURFER_HOME=$(pwd)"
           } > $GITHUB_OUTPUT
-      - uses: Deep-MI/FastSurfer/.github/actions/build-docker@dev
+      - uses: dkuegler/FastSurfer/.github/actions/build-docker@dev
```
m-reuter:
```suggestion
      - uses: Deep-MI/FastSurfer/.github/actions/build-docker@dev
```

doc/index.rst
```diff
@@ -2,6 +2,8 @@
    sphinx-quickstart on Thu Nov 30 15:48:44 2023.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
+   ---
+   Modified and updated by David Kügler of the FastSurfer Team.
 
    Modified and updated by the David Kügler of the FastSurfer Team.
```
m-reuter:
```suggestion
   Modified and updated by David Kügler of the FastSurfer Team.
```

doc/index.rst
```diff
@@ -2,6 +2,8 @@
    sphinx-quickstart on Thu Nov 30 15:48:44 2023.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
+   ---
+   Modified and updated by David Kügler of the FastSurfer Team.
```
m-reuter:
```suggestion
```

.github/workflows/quicktest.yaml
dkuegler: All changes to this file are not intended. I figure out why. I had to use this to run these in my dev branch, which I based this on instead of upstream

dkuegler: So I will drop commot 1 of 2.

doc/index.rst
dkuegler: This is only a comment in any case...

PR #762, opened 2026-01-08
FastSurfer-CC bugfixes

ClePol:
This PR fixes multiple bugs in the corpus callosum module.

One critical bug was in the AC/PC localization, where the AC/PC was never correctly placed on the midslice.

Other bugs are mainly related to visualization and alternate subdivision schemes (not run in the default run_fastsurfer.sh configuration).

### DIFF
--- a/CorpusCallosum/fastsurfer_cc.py
+++ b/CorpusCallosum/fastsurfer_cc.py
@@ -841,7 +841,7 @@ def main(
     # Process slices based on selection mode
 
     logger.info(f"Processing slices with selection mode: {slice_selection}")
-    slice_results, slice_io_futures, cc_contours, cc_mesh = recon_cc_surf_measures_multi(
+    slice_results, slice_io_futures, cc_contours = recon_cc_surf_measures_multi(
         segmentation=cc_fn_seg_labels,
         slice_selection=slice_selection,
         upright_header=fsavg_header,
@@ -883,8 +883,8 @@ def main(
         if len(middle_slice_result["split_contours"]) <= 5:
             cc_subseg_midslice = make_subdivision_mask(
                 (cc_fn_seg_labels.shape[1], cc_fn_seg_labels.shape[2]),
-                middle_slice_result["split_contours"],
-                vox2ras=fsavg_vox2ras @ np.linalg.inv(fsavg2midslice_vox2vox),
+                middle_slice_result["subdivision_lines"],
+                vox2ras=fsavg_vox2ras @ np.linalg.inv(fsavg2midslice_vox2vox)
             )
         else:
             logger.warning("Too many subsegments for lookup table, skipping sub-division of output segmentation.")
@@ -928,8 +928,8 @@ def main(
         additional_metrics["cc_5mm_volume_pv_corrected"] = cc_volume_contour
 
     # get ac and pc in all spaces
-    ac_coords_3d = np.hstack((FSAVERAGE_MIDDLE, ac_coords_vox))
-    pc_coords_3d = np.hstack((FSAVERAGE_MIDDLE, pc_coords_vox))
+    ac_coords_3d = np.hstack((FSAVERAGE_MIDDLE / vox_size[0], ac_coords_vox))
+    pc_coords_3d = np.hstack((FSAVERAGE_MIDDLE / vox_size[0], pc_coords_vox))
     standardized2orig_vox2vox, ac_coords_standardized, pc_coords_standardized, ac_coords_orig, pc_coords_orig = (
         calc_mapping_to_standard_space(orig, ac_coords_3d, pc_coords_3d, orig2fsavg_vox2vox)
     )
--- a/CorpusCallosum/fastsurfer_cc.py
+++ b/CorpusCallosum/fastsurfer_cc.py
@@ -914,17 +914,19 @@ def main(
 
     ########## Save outputs ##########
     additional_metrics = {}
-    if len(outer_contours) > 1:
-        cc_volume_voxel = segmentation_postprocessing.get_cc_volume_voxel(
+
+    cc_volume_voxel = segmentation_postprocessing.get_cc_volume_voxel(
             desired_width_mm=5,
             cc_mask=np.equal(cc_fn_seg_labels, CC_LABEL),
             voxel_size=vox_size, # in LIA order
         )
+    additional_metrics["cc_5mm_volume"] = cc_volume_voxel
+
+
+    if len(outer_contours) > 1:
         logger.info(f"CC volume voxel: {cc_volume_voxel}")
         cc_volume_contour = calculate_cc_volume_contour(cc_contours, width=5.0)
         logger.info(f"CC volume contour: {cc_volume_contour}")
-
-        additional_metrics["cc_5mm_volume"] = cc_volume_voxel
         additional_metrics["cc_5mm_volume_pv_corrected"] = cc_volume_contour
 
     # get ac and pc in all spaces
--- a/CorpusCallosum/shape/curvature.py
+++ b/CorpusCallosum/shape/curvature.py
@@ -61,7 +61,7 @@ def compute_mean_curvature(path: Points2dType) -> float:
 
 def calculate_curvature_metrics(
     midline: Points2dType,
-    split_points: np.ndarray | None = None,
+    split_points: np.ndarray,
 ) -> tuple[float, float, np.ndarray]:
     """
     Calculate curvature metrics for the CC midline, including overall mean,
@@ -71,10 +71,8 @@ def calculate_curvature_metrics(
     ----------
     midline : Points2dType
         Equidistant points along the midline.
-    split_points : np.ndarray, optional
+    split_points : np.ndarray
         Points on the midline where it was split (for orthogonal subdivision).
-    split_contours : ContourList, optional
-        List of split contours (for other subdivision methods).
 
     Returns
     -------
@@ -95,10 +93,9 @@ def calculate_curvature_metrics(
 
     # Find split indices on the midline for subsegment curvature
     split_indices_midline = [0]
-    if split_points is not None:
-        for sp in split_points:
-            idx = np.argmin(np.linalg.norm(midline - sp, axis=1))
-            split_indices_midline.append(idx)
+    for sp in split_points:
+        idx = np.argmin(np.linalg.norm(midline - sp, axis=1))
+        split_indices_midline.append(idx)
 
     split_indices_midline.append(len(midline) - 1)
     split_indices_midline.sort()
--- a/CorpusCallosum/shape/curvature.py
+++ b/CorpusCallosum/shape/curvature.py
@@ -17,7 +17,7 @@ import numpy as np
 from CorpusCallosum.utils.types import Points2dType
 
 
-def compute_curvature(path: Points2dType) -> np.ndarray[tuple[int], np.dtype[np.float_]]:
+def compute_curvature(path: Points2dType) -> np.ndarray[tuple[int], np.dtype[np.float64]]:
     """Compute curvature by computing edge angles.
 
     Parameters
--- a/CorpusCallosum/shape/curvature.py
+++ b/CorpusCallosum/shape/curvature.py
@@ -14,7 +14,7 @@
 
 import numpy as np
 
-from CorpusCallosum.utils.types import ContourList, Points2dType
+from CorpusCallosum.utils.types import Points2dType
 
 
 def compute_curvature(path: Points2dType) -> np.ndarray[tuple[int], np.dtype[np.float_]]:
@@ -62,7 +62,6 @@ def compute_mean_curvature(path: Points2dType) -> float:
 def calculate_curvature_metrics(
     midline: Points2dType,
     split_points: np.ndarray | None = None,
-    split_contours: ContourList | None = None,
 ) -> tuple[float, float, np.ndarray]:
     """
     Calculate curvature metrics for the CC midline, including overall mean,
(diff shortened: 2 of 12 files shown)


CorpusCallosum/shape/curvature.py
dkuegler:
```suggestion
       Points on the midline where it was split (for orthogonal subdivision).
```

dkuegler: Does this not make `split_points` mandatory?

CorpusCallosum/shape/postprocessing.py
dkuegler: return type
```suggestion
) -> tuple[list[CCMeasuresDict], list[concurrent.futures.Future], list[CCContour]]:
```

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
             for split_contour in split_contours
         ]
+        subdivision_lines = [
+            transform_to_acpc_standard(line.T, *acpc_contour_coords_as)[0].T
+            for line in subdivision_lines
+        ]
     elif subdivision_method == "vertical":
-        areas, split_contours = subdivide_contour(contour_in_acpc_space, subdivisions, plot=False)
+        areas, split_contours, split_points_midline, subdivision_lines = subdivide_contour(
+            contour_in_acpc_space, subdivisions, plot=False
```
dkuegler:
```suggestion
            contour_in_acpc_space, subdivisions, plot=False,
```

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
+        )
     elif subdivision_method == "angular":
         if not np.allclose(np.diff(subdivisions), np.diff(subdivisions)[0]):
             raise ValueError(
                 f"Angular subdivision method (Hampel) only supports equidistant subdivision, "
                 f"but got: {subdivisions}. No measures are computed.",
             )
-        areas, split_contours = hampel_subdivide_contour(contour_in_acpc_space, num_rays=len(subdivisions), plot=False)
+        areas, split_contours, split_points_midline, subdivision_lines = hampel_subdivide_contour(
+            contour_in_acpc_space, num_rays=len(subdivisions), plot=False
```
dkuegler:
```suggestion
            contour_in_acpc_space, num_rays=len(subdivisions), plot=False,
```

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
+        )
     elif subdivision_method == "eigenvector":
         pt0, pt1 = get_primary_eigenvector(contour_in_acpc_space)
         contour_eigen, _, _, rotate_back_eigen = transform_to_acpc_standard(contour_in_acpc_space, pt0, pt1)
         ac_pt_eigen, _, _, _ = transform_to_acpc_standard(ac_pt_acpc[:, None], pt0, pt1)
         ac_pt_eigen = ac_pt_eigen[:, 0]
-        areas, split_contours = subdivide_contour(contour_eigen, subdivisions, oriented=True, hline_anchor=ac_pt_eigen)
+        areas, split_contours, split_points_midline, subdivision_lines = subdivide_contour(
+            contour_eigen, subdivisions, oriented=True, hline_anchor=ac_pt_eigen
```
dkuegler:
```suggestion
            contour_eigen, subdivisions, oriented=True, hline_anchor=ac_pt_eigen,
```

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
 
     # Apply different subdivision methods based on user choice
     split_contours: ContourList
+    subdivision_lines: list[Points2dType]
     split_points_midline: np.ndarray | None = None
     if subdivision_method == "shape":
```
dkuegler: This whole if block should probably be pulled out into a function, so the function contract (input/output/etc) is also documented better... and this would separate it from the code around it.

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
+        # Transform from the outputs back to the input space
         split_contours = [rotate_back_eigen(split_contour) for split_contour in split_contours]
+        subdivision_lines = [rotate_back_eigen(line.T).T for line in subdivision_lines]
+        split_points_midline = rotate_back_eigen(np.asarray(split_points_midline).T).T
     else:
         raise ValueError(f"Invalid subdivision method {subdivision_method}")
     
-    # order areas anterior to posterior
-    areas = areas[::-1]
+    # NOTE: areas, subdivision_lines, and split_points_midline are all ordered Anterior to Posterior
```
dkuegler: This comment would be part of the function docstring (see above).

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
 
     total_area = np.sum(areas)
-    # total_perimeter should include the edge from last to first point
     contour_closed = np.concatenate([contour_as, contour_as[:, :1]], axis=1)
     total_perimeter = np.sum(np.linalg.norm(np.diff(contour_closed, axis=1), axis=0))
     circularity = 4 * np.pi * total_area / (total_perimeter**2)
 
-    # Transform split contours back to original space
+    # Transform split contours back to original space (from ACPC to RAS)
```
dkuegler, line 461: If we transform `subdivision_lines` and `split_contours` back to ras here anyways, why not just have the if-block directly return the ras coordinates? seems to be unnecessarily transforming back and forth.

Especially, if shape is not consistent anyways...

CorpusCallosum/shape/postprocessing.py
```diff
@@ -384,50 +381,68 @@ def recon_cc_surf_measure(
...
+    # while the others are in ACPC and need rotate_back_acpc.
     split_contours = [rotate_back_acpc(split_contour) for split_contour in split_contours]
+    subdivision_lines = [rotate_back_acpc(line.T).T for line in subdivision_lines]
+    if subdivision_method != "shape":
+        split_points_midline = rotate_back_acpc(np.asarray(split_points_midline).T).T
 
     # Calculate curvature metrics
     curvature, curvature_body, curvature_subsegments = calculate_curvature_metrics(
-        midline_equi, split_points=split_points_midline, split_contours=split_contours
+        midline_equi, split_points=split_points_midline
```
dkuegler, line 471:
```suggestion
        midline_equi, split_points=split_points_midline,
```

CorpusCallosum/shape/postprocessing.py
dkuegler:
```suggestion
    """Test whether points in coords are to the left of the line (line_start->line_end).
```

CorpusCallosum/shape/subsegment_contour.py
dkuegler, line 156: This is missing the documentation for the last return value.

CorpusCallosum/shape/subsegment_contour.py
```diff
@@ -339,50 +339,14 @@ def subsegment_midline_orthogonal(
...
-    first_contour = split_contours[0]
-    # Check each point against all other contours
-    for contour in split_contours[1:]:
-        # 0: coord-axis, 1: contour-axis, 2: first_contour_axis
-        contour_comparison = np.isclose(first_contour[:, None], contour[:, :, None], atol=1e-6)
-        # mask of contour points, that are also in first_contour (axis 1 after all)
-        contour_points_in_first_contour_mask = np.any(np.all(contour_comparison, axis=0), axis=1)
-        unique_contour_points.append(contour[:, ~contour_points_in_first_contour_mask].T)
+    # add original contour to split_contours to be the union of all subsegments
+    split_contours.append(contour)
```
dkuegler, line 353: Does that mean the standardization of what `split_contours` means has changed? as it used to have the full contour on index 0, now on -1?

CorpusCallosum/shape/subsegment_contour.py
dkuegler, line 378: Missing documentation for last 2 return values.

dkuegler:
```suggestion
            raise ValueError("No intersections found, is contour valid?")
```

dkuegler, line 581: Missing documentation of 2 return values

CorpusCallosum/shape/subsegment_contour.py
```diff
@@ -717,8 +695,8 @@ def subdivide_contour(
                 )
             )
 
-            # add cumulative subsegment
-            split_contours.append(start_to_cutoff)
+            # add cumulative subsegment (Anterior -> Current Split)
+            split_contours.append(anterior_to_cutoff)
         else:
             raise ValueError("No intersections found, this should not happen")
```
dkuegler:
```suggestion
            raise ValueError("No intersections found, is contour valid?")
```

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -928,8 +928,8 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
         additional_metrics["cc_5mm_volume_pv_corrected"] = cc_volume_contour
 
     # get ac and pc in all spaces
-    ac_coords_3d = np.hstack((FSAVERAGE_MIDDLE, ac_coords_vox))
-    pc_coords_3d = np.hstack((FSAVERAGE_MIDDLE, pc_coords_vox))
+    ac_coords_3d = np.hstack((FSAVERAGE_MIDDLE / vox_size[0], ac_coords_vox))
+    pc_coords_3d = np.hstack((FSAVERAGE_MIDDLE / vox_size[0], pc_coords_vox))
     standardized2orig_vox2vox, ac_coords_standardized, pc_coords_standardized, ac_coords_orig, pc_coords_orig = (
         calc_mapping_to_standard_space(orig, ac_coords_3d, pc_coords_3d, orig2fsavg_vox2vox)
     )
```
dkuegler, line 937: This should be correct, but I would prefer, if we used 0 and `fsavg2midslice_vox2vox` then we don't have conversions in as many places and it is more consistent.

dkuegler (commented):
Minor changes.

CorpusCallosum/shape/postprocessing.py
ClePol: Why is shape not consistent? We need the subdivisions and split contours in the original space later for visualization, but we never need them in AC/PC standard space

CorpusCallosum/shape/curvature.py
ClePol: yes, changed

CorpusCallosum/shape/postprocessing.py
ClePol: Refactored

CorpusCallosum/shape/subsegment_contour.py
ClePol: Yes. Before it was inconsistent, and the default subdivision scheme placed the full contour at zero. Additionally, some subdivision schemes did subdivision from posterior to anterior, others from anterior to posterior.
Now all should go anterior->posterior, from initial subsegment to the unions of subsegments, ending with the whole contour.

CorpusCallosum/fastsurfer_cc.py
ClePol: I don't know how you organized the transform.
If this is important could you change it change it and confirm the result stays the same (numerically)

CorpusCallosum/shape/subsegment_contour.py
dkuegler: Thanks for clarifying.

CorpusCallosum/fastsurfer_cc.py
dkuegler: I will have a look later!

CorpusCallosum/shape/postprocessing.py
dkuegler: Some parts of the if blocks had transformations of these from ras to standard, just to transform them back here. Other did not. So the functions were not consistent, the the if blocks seems to have been.

ClePol: Ah, I think I understand.
There is a design decision here whether we want to apply all subdivision schemes in standardized space or not. In original publications of reference sub-division schemes, the orientation was not standardized. I think our previous implementation followed this original definition of applying the subdivision method without explicit standardization. This is why the space was transformed individually according to the subdivision scheme (in the if block).

I could show that standardizing the space gives better scores in CC index for example. So I think the current implementation improves the old subdivision schemes in a reasonable way. Additionally, we are not really working in image space anyway because our image is already aligned to fsaverage. So I'm in favor of keeping it as is and applying all subdivision schemes in standardized space.

I hope this is what you meant.

dkuegler: That is fine.  But since our convention is to end up in fsavg space, after the if block (or rather the function that this if-block should be refactored to) we should be in fsavg space. Right now, the convention is to end in acpc_standard.
So after that there is a transformation back to fsavg, while some methods had transformed from fsavg to acpc_standard just before, incurring double transformation.

ClePol: I don't see this issue in the current code
From my view we do 
1. fsavg to acpc
2. subdivision (input and output are in same space)
3. acpc to fsavg

PR #765, opened 2026-01-19
Various updates and changes to the documentation

dkuegler:
This PR has various changes and updates to the fastsurfer documentation.

Two example:
- consistently mount/bind shared folders for singularity and docker in the same location as on the host file system (to simplify the documentation)
- move the singularity example into position 1 (instead of the docker example)

Todos:
- ~~rebase the changes onto dev (currently on ClePol/CC_doc)~~
- ~~review and test the documentation for consistency and completeness~~


stools.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# script for functions used by srun_fastsurfer.sh and srun_freesufer.sh
+# script for functions used by srun_fastsurfer.sh and brun_freesufer.sh
```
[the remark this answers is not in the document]
dkuegler: this is fine as is. Not sure which naming conventions is not followed.

m-reuter:
I fixed the doc build in #772 by adding an exclude

```
numpydoc_validation_exclude = {
    r'\.WarmupCosineLR\.step$',  # Exclude only WarmupCosineLR.step
}
```
to ```conf.py``` .

CerebNet/utils/lr_scheduler.py
```diff
@@ -133,7 +133,7 @@ def _reset_lr(self):
 
 
 # https://detectron2.readthedocs.io/_modules/detectron2/solver/lr_scheduler.html
-class WarmupCosineLR(torch.optim.lr_scheduler._LRScheduler):
+class WarmupCosineLR(torch.optim.lr_scheduler._LRScheduler):  # numpydoc ignore=GL09
```
m-reuter:
```suggestion
class WarmupCosineLR(torch.optim.lr_scheduler._LRScheduler):
```

dkuegler: Resolved, but this is really necessary because the underlying error originates in the torch base class. I guess this information could be added in a comment in the code, but the doc does not build without this comment.

dkuegler: Resolved, but this is really necessary because the underlying error originates in the torch base class. I guess this information could be added in a comment in the code, but the doc does not build without this comment.

m-reuter: yes, this was fixed in #772 , not sure what you mean. This comment which I reversed, did not solve it.

dkuegler:
Well, I guess that is also a way to fix the doc build error ...

m-reuter:
The local comment did not fix it, which is why I tried fixing it in conf.py.

PR #766, opened 2026-01-19
bump python packages, use UV instead of conda, update docker build

dkuegler:
This PR includes:
- version bumps for all python packages
- removes conda environment files entirely (fastsurfer now uses UV as the default/supported package manager in both documentation and docker image)
- several updates to the docker build
  - UV instead of docker (significant simplification of the build)
  - update to Ubuntu 24.04 as base
  - update all major python packages
  - add xvfb-run dependencies in the runtime to support whippersnappy (for example for CC QC images)
  - add image metadata fields
  - adaptations to the build script
  - caching the freesurfer download instead of downloading each time is a massive speedup to the build time
  - fspython is a wrapper script instead of a softlink

Todo:
- ~~finish surface pipeline testing~~
- ~~currently there is an error where Freesurfer's `rca-config` (a python script) cannot import `yaml` but yaml is installed and available in the distributed python environment~~


run_fastsurfer.sh
```diff
@@ -738,6 +738,33 @@ then
...
+  # if we have xvfb-run, we can use it to provide a virtual display
+  if [[ -n "$(which xvfb-run)" ]] ; then maybe_xvfb=("xvfb-run" "-a") ; fi
+
+  # try loading opengl, if this is successful we are fine
+  py_opengltest="import sys ; import glfw ; import whippersnappy.core ; sys.exit(1-glfw.init())"
+  opengl_error_message="$("${maybe_xvfb[@]}" $python -c "$py_opengltest" 2>&1 > /dev/null)"
+  if [[ "$?" != "0" ]]
+  then
+    # if we cannot import OpenGL or whippersnappy, its an environment installation issue
+    if [[ "$opengl_error_message" =~ "ModuleNotFoundError" ]] || [[ "$opengl_error_message" =~ "ImportError" ]]
```
m-reuter, line 754: I would produce a Warning and skipt the QC plots that require OpenGL. I would not exit with an ERROR here.

dkuegler: updated to this behavior (needs to be tested)

tools/Docker/Dockerfile
```diff
@@ -170,55 +280,49 @@ ENV OS=Linux \
...
+source /venv/bin/activate
+# Fix for cuda11.8+cudnn8.7 bug+warning: https://github.com/pytorch/pytorch/issues/97041
+if [[ "$DEVICE" == "cu118" ]]
+then
+  python3 <<CU118FIX
+from pathlib import Path
+import torch
+
+p = Path(torch.__file__).parent / "lib"
+(p / "libnvrtc.so").symlink_to(p / p.glob("libnvrtc-*.so.11.2"))
```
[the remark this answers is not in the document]
dkuegler: fixed

PR #768, opened 2026-01-21
Quick fix to address warning in run_prediction.py

dkuegler:
This PR fixes the following warning by manually converting to an ndarray
```
/fastsurfer/FastSurferCNN/run_prediction.py:243: UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at /pytorch/torch/csrc/utils/tensor_numpy.cpp:203.)
```


dkuegler:
The doc error is coming from somewhere else and should not be fixed in this PR, fixed in #765 .

Issue #699, opened 2025-05-27
Ipynb colab notebook: FastSurfer asegdkt segmentation failed


razoralm:
I am testing your colab Notebook 
[Complete_FastSurfer_Tutorial.ipynb](https://github.com/deep-mi/FastSurfer/tree/stable/Tutorial#notebook-2---complete-fastsurfer-tutorial)

The first step - 3 clicks 
I want to do in with my MRI data (the samen .mgz package, 360 saggital T1 slices) 

After running the ERROR is
```
 /content/fastsurfer//run_fastsurfer.sh: line 1096:  5931 Killed                  "${cmd[@]}"
ERROR: FastSurfer asegdkt segmentation failed.
```

dkuegler:
Hello, thank you for creating an issue here.
Please share more of the logfile/console output.

It looks to me like the run was terminated, before it could finish "naturally". But it is unclear by what it was stopped.

m-reuter:
For me the 3rd step (run_fastsurfer) works and reports the total volume of the segmentation. But then it get's stuck (wheel turns, I waited 9 minutes). Stopping the cell works and checking the final segmentation looks fine.

m-reuter:
Tested it again and everything works fine and it hangs in the run_fastsurfer.sh command after it completes all steps. The cell can then be stopped and the results downloaded or looked at. 

@dkuegler we should test (and fix) the ipynb script on colab with the new RC before we release it. 
I debugged this a little more today and both run_prediction.py terminates successfully and also run_fastsurfer.sh reaches the end, but the cell does not stop. I tried closing all files, and killing all sub-processes both in the shell and in the python scripts, but so far no success.

m-reuter:
Found the issue: 
version.py hangs on google collab with the ```--sections all``` . One solution would be to do ```--sections "+git"```only. The other is to set a timeout when sending it to background. I will do the later as a quick fix.

PR #774, opened 2026-01-25
Add timeout to version.py execution

m-reuter:
version.py with ```--sections all```hangs on google collab. see #699 

Possible solutions:
- to not call all section, but e.g. only "+git"
- to fix whatever is hanging in version.py (if that is even possible as maybe the pip stuff is too slow on collab or the md5 computation)
- put a timeout to ensure that the background process is terminated eventually (this is what I do here as a quick fix)


dkuegler:
There are technically 3 components to version, pip, git, and checkpoints. ... I think instead of adding a timeout to the version call altogether. It would be better to add timeouts to the individual components internally (as those are mostly calls to binaries, e.g.. git, md5 or pip).

This version would not generate the build file at all I believe, but incomplete build files should be much preferred...

PR #781, opened 2026-02-04
Fixes to the docker build process

dkuegler:
This PR addresses four issues, three of which are related to docker:

1. uv properly installs the cuda libraries into the environment, so the CUDA 11.8 fix to link the cuda libraries is no longer required
2. Expose the --insecure flag from `install_fs_pruned.sh` through the Docker build-args to `build.py` and its commandline arguments
3. Change the way the repository_url annotation is read from based on the git remote the current branch is tracking to reading it from pyproject.toml. This fixes the build issue in the quicktest build.
4. Fix the warning that pops up for non-writable numpy array.

### Todo
- ~~test whether issue 4 is finally fixed~~

### DIFF
--- a/.github/workflows/quicktest.yaml
+++ b/.github/workflows/quicktest.yaml
@@ -24,7 +24,7 @@ concurrency:
 
 on:
   pull_request:
-    types: [labeled, unlabeled]
+    types: [labeled]
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


dkuegler:
Both issues raised by copilot have been resolved.

Issue #780, opened 2026-02-03
Is the longitudinal pipeline primarily running on GPU or CPU?


RDoerfel:
Dear FastSurfer team, 

I saw that FastSurfer supports longitudinal processing now. Pretty cool! 

I was wondering is it primarily running on GPU or CPU? I would like to parallise it to run multiple subjects in parallel and I'm wondering whether I should run on our CPU cluster (no GPU), or GPU cluster (rather little CPU). 

Best, Ruben

dkuegler:
It uses both in concert. We already have a specialized script that [schedules the CPU and GPU-heavy parts of **cross-sectional** FastSurfer on a SLURM cluster](https://deep-mi.org/FastSurfer/dev/scripts/SLURM.html).

For the longitudinal pipeline, this is different and it is a mix of GPU and CPU workloads across multiple steps.

Thus far, there has not been any demand/requests to extend the cluster scheduling script to longitudinal processing, but you should be able to orchestrate the calls yourself as well.

[The longitudinal processing](https://deep-mi.org/FastSurfer/dev/overview/LONG.html#behind-the-scenes) is laid out in `long_fastsurfer.sh`.

`long_fastsurfer.sh` already has a `--parallel` flag (and `--parallel_seg` and `--parallel_surf` subflags, see [the documentation](https://deep-mi.org/FastSurfer/dev/scripts/long_fastsurfer.html)).

If you take a look at [the longitudinal processing](https://deep-mi.org/FastSurfer/dev/overview/LONG.html#behind-the-scenes), it has 5 components:

1. Template Init
2. Template Seg
3. Template Surf
4. Long Seg
5. Long Surf

Steps 1, 2, 3 are GPU-heavy and Step 3 and 5 are CPU-heavy.

So the simplest way to approach to achieve the distribution of the jobs would be to 

I. Copy `long_fastsurfer.sh` (`long_fastsurfer_CPU-heavy.sh`) and comment out steps 3 and 5, then run the script on your GPU cluster (using all other parameters as you would use).

II. Copy `long_fastsurfer.sh` (`long_fastsurfer_GPU-heavy.sh`) and comment out steps 1, 2, 4, the run the script on your CPU cluster (using the same parameters as in I.)

I have not tried this myself, but from a quick inspection of `long_fastsurfer.sh` this should work with no further edits.

RDoerfel:
cool, thanks!

RDoerfel:
[...]

PR #786, opened 2026-02-12
add stages to long_fastsurfer.sh

m-reuter:
Enhancement requested #780 

## Summary

Adds stage-based execution to `long_fastsurfer.sh`, allowing users to run individual pipeline stages independently for flexible resource allocation across compute nodes with varying GPU/CPU capabilities.

## Changes

- Added `--stage` flag to select which stages to run: `prepare`, `template_seg`, `template_surf`, `long_seg`, `long_surf`, or `all` (default)
- Automatic dependency validation ensures required stages have completed before running dependent stages
- Smart input validation: only requires `--t1s` when running `prepare` stage
- Backward compatible: default behavior unchanged (runs all stages)

## Usage

```bash
# Run full pipeline (default, unchanged)
./long_fastsurfer.sh --tid template01 --t1s tp1.mgz tp2.mgz --tpids tp1 tp2

# Run individual stage
./long_fastsurfer.sh --tid template01 --tpids tp1 tp2 --stage template_seg

# Run multiple stages
./long_fastsurfer.sh --tid template01 --tpids tp1 tp2 --stage template_surf --stage long_surf
```

## Benefits

- **Resource optimization**: Send GPU stages to GPU nodes, CPU stages to CPU nodes
- **Restart capability**: Re-run individual stages without reprocessing completed ones
- **Better cluster scheduling**: Split heavy jobs into smaller, faster-to-schedule jobs

### DIFF
--- a/long_fastsurfer.sh
+++ b/long_fastsurfer.sh
@@ -317,7 +317,23 @@ function check_stage_dependencies() {
 # Warning if run as root user
 check_allow_root "${brun_flags[@]}" # --allow_root must be passed to brun
 
-# Check if prepare stage will run - only then we need t1s and tpids
+# Validate required inputs based on which stages will run
+if [ -z "$tid" ]
+ then
+  echo "ERROR: Must supply person-specific template name via --tid <template id>!"
+  exit 1
+fi
+
+# tpids required for prepare, long_seg, and long_surf stages
+if should_run_stage "prepare" || should_run_stage "long_seg" || should_run_stage "long_surf"; then
+  if [ "${#tpids[@]}" -lt 1 ]
+   then
+    echo "ERROR: Must supply time point IDs via --tpids <timepoint id 1> [<timepoint id 2> ...]!"
+    exit 1
+  fi
+fi
+
+# t1s only required for prepare stage
 if should_run_stage "prepare"; then
   if [ "${#t1s[@]}" -lt 1 ]
    then
@@ -325,33 +341,12 @@ if should_run_stage "prepare"; then
     exit 1
   fi
 
-  if [ "${#tpids[@]}" -lt 1 ]
-   then
-    echo "ERROR: Must supply time points ids via --tpids <timepoint id 1> [<timepoint id 2> ...] when running 'prepare' stage!"
-    exit 1
-  fi
-
-  # check that t1s list is same length as tpids
+  # For prepare stage, t1s and tpids must have same length
   if [ "${#tpids[@]}" -ne "${#t1s[@]}" ]
    then
     echo "ERROR: Length of tpids must equal t1s!"
     exit 1
   fi
-else
-  # For other stages, we still need tpids but not necessarily t1s
-  if should_run_stage "long_seg" || should_run_stage "long_surf"; then
-    if [ "${#tpids[@]}" -lt 1 ]
-     then
-      echo "ERROR: Must supply time points ids via --tpids <timepoint id 1> [<timepoint id 2> ...] when running 'long_seg' or 'long_surf' stages!"
-      exit 1
-    fi
-  fi
-fi
-
-if [ -z "$tid" ]
- then
-  echo "ERROR: Must supply person-specific template name via --tid <template id>!"
-  exit 1
 fi
 
 # check that SUBJECTS_DIR exists
--- a/long_fastsurfer.sh
+++ b/long_fastsurfer.sh
@@ -59,6 +59,10 @@ LF=""
 brun_flags=()
 python="python3 -s" # avoid user-directory package inclusion
 
+# Stage control
+run_stages=("prepare" "template_seg" "template_surf" "long_seg" "long_surf")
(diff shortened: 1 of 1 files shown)


long_fastsurfer.sh
```diff
@@ -194,19 +233,134 @@ done
 
 source "${reconsurfdir}/functions.sh"
 
+# Function to check if a stage should run
+function should_run_stage() {
+  local stage=$1
+  for s in "${run_stages[@]}"; do
+    if [[ "$s" == "$stage" ]]; then return 0; fi
```
dkuegler, line 240: This should allow `should_run_stage "prepare|template_seg"`

```suggestion
    IFS="|"
    for this_stage in $stage; do
      if [[ "$s" == "$this_stage" ]]; then return 0; fi
    done
```

long_fastsurfer.sh
```diff
@@ -194,19 +233,134 @@ done
...
+              echo "ERROR: Stage '$stage' requires 'prepare' long-inputs for ALL time points."
+              echo "  Missing long-inputs for: ${missing_long_inputs[*]}"
+              exit 1
+            fi
+          fi
+          # Only check once, then break
+          break
+          ;;
+      esac
+    done
```
dkuegler: Is this not everything except prepare? I guess this is not too important, but the logic is somewhat convoluted here... why not just use your `should_run_stage` function to unravel this here?

```suggestion
     if should_run_stage "template_seg|template_surf|long_seg|long_surf"; then
       # Basic checks: template directory and files from 'prepare'
       if [[ ! -d "$sd/$tid" ]] || [[ ! -f "$sd/$tid/mri/orig.mgz" ]]; then
         echo "ERROR: Stage '$stage' requires 'prepare' to have been run first."
         echo "  Template directory $sd/$tid or $sd/$tid/mri/orig.mgz not found."
         exit 1
       fi
       # Additional check: template must be fully prepared for --base/--long runs
       if [[ ! -f "$sd/$tid/base-tps.fastsurfer" ]]; then
         echo "ERROR: Stage '$stage' requires 'prepare' to have been fully completed."
         echo "  Template indicator file $sd/$tid/base-tps.fastsurfer not found."
         exit 1
       fi
    fi
    # Long stages additionally require long-inputs for all time points
    if should_run_stage "long_seg|long_surf"; then
       missing_long_inputs=()
       for tpid in "${tpids[@]}"; do
          if [[ ! -f "$sd/$tid/long-inputs/$tpid/long_conform.nii.gz" ]]; then
             missing_long_inputs+=("$tpid")
          fi
       done
       if [[ ${#missing_long_inputs[@]} -gt 0 ]]; then
          echo "ERROR: Stage '$stage' requires 'prepare' long-inputs for ALL time points."
          echo "  Missing long-inputs for: ${missing_long_inputs[*]}"
          exit 1
       fi
    fi
```

long_fastsurfer.sh
```diff
@@ -241,132 +388,160 @@ function log () { echo "$1" | tee -a "$LF" ; }
...
+      date
+    } > "$long_seg_cmdf_log"
+    echo "#/bin/bash" > "$long_seg_cmdf"
+    run_it_cmdf "$LF" "$long_seg_cmdf" "${cmda[@]}"
+    log "Starting longitudinal segmentations, logs temporarily diverted to $long_seg_cmdf_log..."
+    log "Output from this process will be delayed to when it has finished."
+    log "======================================="
+    # at the end of the job below, the gpu can be released (for tight management of resources, run
+    # Surfaces in different jobs. Alternative, add a command to "$long_seg_cmdf" that releases the gpu or
+    # triggers the next "subject"
```
dkuegler: I guess this comment is obsolete now with the stages
```suggestion
```

long_fastsurfer.sh
```diff
@@ -241,132 +388,160 @@ function log () { echo "$1" | tee -a "$LF" ; }
...
+      log "Terminating because ${what_failed[*]} failed!"
+      exit 1
+    fi
   fi
+  run_it "$LF" "${cmda[@]}"
 fi
-run_it "$LF" "${cmda[@]}"
 
 log "======================================="
 log "Full longitudinal processing for $tid finished!"
```
dkuegler: If we only did a specific stage, the log should not say "Full longitudinal processing ... finished"

```suggestion
if should_run_stage "long_surf"; then what="Full"; else what="Stage(s) ${run_stages[*]}"; fi
log "$what longitudinal processing for $tid finished!"
```

dkuegler (commented):
I think I would use should_run_stage also in the validation check function.
And at the end only print "Full" if it is actually the final stage.

long_fastsurfer.sh
m-reuter: I vote for keeping this function simple to only check one stage at a time and instead use the standard bash || in the if statements.

long_fastsurfer.sh
m-reuter: yes, this and some code further down could be simplified a bit.

dkuegler (approved):
Looks good to me 👍

PR #788, opened 2026-02-13
Fix bug in spherically project wrapper

dkuegler:
This PR fixes two bugs in the use of spherically_project_wrapper.py. Both of them only appear if the baseline python implementation fails.

1. the call to `spherically_project_wrapper.py` uses a `--sd` argument, but the value was previously passed as empty (the used variable did not exist).
2. There was a missing function in `MessageBuffer`: `forward_output`, which forwards stdout and stderr content through the wrapper to the python stdout. This function missing raised an error.

Additionally, the PR fixes a near-irrelevant signature-inconsistency between the `run_tools` extension of `Popen.__init__` and the underlying `subprocess.Popen.__init__`, where `subprocess.Popen` only accepts one positional parameter, but `run_tools.Popen` accepted multiple ... only to then pass that to `subprocess.Popen`, where it would be invalid.

Todo:
- ~~test in bigger dataset to cover both spherically project implementations~~

### DIFF
--- a/FastSurferCNN/utils/run_tools.py
+++ b/FastSurferCNN/utils/run_tools.py
@@ -64,6 +64,35 @@ class MessageBuffer:
     def err_str(self, encoding="utf-8"):
         return self.err.decode(encoding=encoding)
 
+    def forward_output(
+        self,
+        file: io.TextIOBase | None = None,
+        encoding: str | None = None,
+        out_prefix: str = "",
+        err_prefix: str = "!",
+    ):
+        """
+        Forward the content of this MessageBuffer to a file or stream.
+
+        Parameters
+        ----------
+        file : IO.TextIO, optional
+            The file or stream to which the output should be forwarded (defaults to stdout).
+        encoding : str, optional
+            Charset to encode.
+        out_prefix : str, default=""
+            String to prefix lines from the stdout output.
+        err_prefix : str, default="!"
+            String to prefix lines from the stderr output.
+        """
+        if file is None:
+            from sys import stdout
+            file = stdout
+        for line in self.out_str(encoding=encoding).splitlines():
+            print(f"{out_prefix}{line}", file=file)
+        for line in self.err_str(encoding=encoding).splitlines():
+            print(f"{err_prefix}{line}", file=file)
+
 
 class Popen(subprocess.Popen):
     """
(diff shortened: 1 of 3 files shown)


recon_surf/recon-surf.sh
```diff
@@ -775,7 +775,7 @@ for hemi in lh rh ; do
       RunIt "$cmd" "$LF" "$CMDF"
     else
       # instead of mris_sphere, directly project to sphere with spectral approach equivalent to -qsphere (23sec)
-      cmda=("${binpath}spherically_project_wrapper.py" --hemi "$hemi" --sdir "$sdir" --subject "$subject")
+      cmda=("${binpath}spherically_project_wrapper.py" --hemi "$hemi" --sdir "$SUBJECTS_DIR" --subject "$subject")
```
m-reuter: actually here the spherically_project_wrapper help text says it expects the surface dir inside the subjects directory, so it was correct before ? 

    parser.add_argument("--sdir", type=Path, help="Surface directory of subject.")

m-reuter: actually that help text needs fixing, as it seems to really expect SUBJECTS_DIR

https://github.com/Deep-MI/FastSurfer/blob/61e80d3665f7f328f0b6258aad7cf9c3c7c50c4c/recon_surf/spherically_project_wrapper.py#L91

dkuegler: I am not really sure why this was working.... so the issue is that the FreeSurfer binary definitly needed the `SUBJECTS_DIR`, but it seems now that your function may need the surface dir...?

dkuegler: I shall fix this to properly use `SUBJECTS_DIR` as recon-all seems to require and thus make it consistent.

m-reuter: yes, other places want the surface dir

https://github.com/Deep-MI/FastSurfer/blob/61e80d3665f7f328f0b6258aad7cf9c3c7c50c4c/recon_surf/spherically_project_wrapper.py#L59

dkuegler: c363190 fixes this inconsistency.
I will run the full test battery over this once more, so we are sure this actually works.

FastSurferCNN/utils/run_tools.py
```diff
@@ -64,16 +64,45 @@ def out_str(self, encoding="utf-8"):
...
     """
     Extension of subprocess.Popen for convenience.
     """
     _starttime: datetime | None = None
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, args, /, **kwargs):
         self._starttime = datetime.now()
-        super().__init__(*args, **kwargs)
+        super().__init__(args, **kwargs)
```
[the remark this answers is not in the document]
dkuegler: Addressed in 2ea6bef

PR #791, opened 2026-02-19
CC module bugfix and robustness

ClePol:
This PR contains two changes:
1. The fsaverage header saved as json file was incorrect, leading to faulty .ltas. This did not affect the segmentations, only applying the .lta with e.g. mri_convert.
2. When the segmentation of the CC was irregular, e.g. due to segmentation failure or (partial) CC agenesis, the morphometry raised errors (as intended). Unfortunately, this meant that the whole FastSurfer pipeline crashed. This update handles morphometry errors more gracefully by still writing the CC segmentation and just reporting "null" for the morphometry of CC slices that could not be processed. Errors are still logged.

I also included an additional check in the SubjectDirectory classes path resolution that prevents writing the full path to the subject folder into the subject folder itself. I have encountered this multiple times when running locally and should not have effects when running in Docker. Feel free to revert if unintended.


ClePol:
Converted to draft, because the orient_volume.lta seems to be not limited to inplane rotation

ClePol:
commited a fix for the orient volume lta, everything seems fine now

m-reuter:
please fix all automated checks before marking it as ready for review.

FastSurferCNN/utils/common.py
```diff
@@ -255,10 +255,21 @@ def filename_in_subject_folder(self, filepath: str | Path) -> Path:
         Path
             The path to the file in the subject folder.
         """
-        if Path(filepath).is_absolute():
-            return Path(filepath)
-        else:
-            return self.subject_dir / self._id / filepath
+        path = Path(filepath)
```
dkuegler: I am voting against this change.

the API asks for either relative paths to subjects_dir or absolute paths...
this introduces relative paths to CWD, I don't want this.

To me this is a confusing and potentially error-prone approach. (ambiguity of relative paths).

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -902,12 +920,15 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
     metrics: tuple[CCMeasures] = get_args(CCMeasures)
 
     # Record key metrics for middle slice
-    output_metrics_middle_slice = {metric: middle_slice_result[metric] for metric in metrics}
+    if middle_slice_result:
+        output_metrics_middle_slice = {metric: middle_slice_result[metric] for metric in metrics}
+    else:
+        output_metrics_middle_slice = {}
```
dkuegler: Should we maybe print a warning here, if middle_slice_result is None.

I feel it would be good to commit this as a warning to the logfile (that processing a slice did not give a results)

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -974,13 +995,15 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
...
-    if sd.has_attribute("cc_mid_measures"):
+    if sd.has_attribute("cc_mid_measures") and middle_slice_result is not None:
         sd.filename_by_attribute('cc_mid_measures').parent.mkdir(exist_ok=True, parents=True)
         io_futures.append(thread_executor().submit(
             save_cc_measures_json,
             sd.filename_by_attribute('cc_mid_measures'),
             output_metrics_middle_slice | additional_metrics,
         ))
+    elif sd.has_attribute("cc_mid_measures"):
+        logger.warning("Midslice morphometry failed, skipping creation of midslice markers file.")
```
dkuegler: And if the create the warning where it rises, this is also obsolete.

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -1008,16 +1037,17 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
     if sd.has_attribute("cc_orient_volume_lta"):
         sd.filename_by_attribute("cc_orient_volume_lta").parent.mkdir(exist_ok=True, parents=True)
         # save lta to standardized space (fsaverage + nodding + ac to center)
-        orig2standardized_ras2ras = orig.affine @ np.linalg.inv(standardized2orig_vox2vox) @ np.linalg.inv(orig.affine)
+        orig2standardized_ras2ras = fsavg_vox2ras @ \
+            np.linalg.inv(standardized2orig_vox2vox) @ np.linalg.inv(orig.affine)
```
dkuegler: I guess this is the fix to the lta issue we knew about, should we also rename the variable for clarity? because it seems to me it uses fsavg instead of orig now... In fact, this looks like a consistent edit that changed the lta consistently in both the ras2ras and the second file in the lta.
```suggestion
        fsavg2standardized_ras2ras = fsavg_vox2ras @ \
            np.linalg.inv(fsavg2standardized_ras2ras) @ np.linalg.inv(orig.affine)
```

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -920,11 +946,12 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
...
 
-    if len(outer_contours) > 1:
+    if len(valid_cc_contours) > 1:
         logger.info(f"CC volume voxel: {cc_volume_voxel}")
-        cc_volume_contour = calculate_cc_volume_contour(cc_contours, width=5.0)
+        cc_volume_contour = calculate_cc_volume_contour(valid_cc_contours, width=5.0)
         logger.info(f"CC volume contour: {cc_volume_contour}")
-        additional_metrics["cc_5mm_volume_pv_corrected"] = cc_volume_contour
+
+    additional_metrics["cc_5mm_volume_pv_corrected"] = cc_volume_contour
```
dkuegler: Should we change the name of this metric as well? Because this is not pv corrected, it is just a surface-based metric.

I still maintain that the 5mm in the nane for the volume is also not relevant only a 5mm assumption allows us to compute a volume.

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -920,11 +946,12 @@ def _orig2midslab_vox2vox(additional_context: int = 0) -> AffineMatrix4x4:
         )
     additional_metrics["cc_5mm_volume"] = cc_volume_voxel
 
-    if len(outer_contours) > 1:
+    if len(valid_cc_contours) > 1:
         logger.info(f"CC volume voxel: {cc_volume_voxel}")
-        cc_volume_contour = calculate_cc_volume_contour(cc_contours, width=5.0)
+        cc_volume_contour = calculate_cc_volume_contour(valid_cc_contours, width=5.0)
```
dkuegler, line 969: If we are only computing the volume for the valid contours here, does that not invalidate this metric altogether? if a slice is missing, we basically get a approx. 20% reduction ... but no clear indication of what is the reason?

CorpusCallosum/fastsurfer_cc.py
```diff
@@ -737,10 +737,10 @@ def main(
     _aseg_fut = thread_executor().submit(nib.load, sd.filename_by_attribute("aseg_name"))
     orig = cast(nibabelImage, nib.load(sd.conf_name))
 
-    # check that the image is conformed, i.e. isotropic 1mm voxels, 256^3 size, LIA orientation
-    if not is_conform(orig, vox_size=None, img_size=None, orientation=None):
+    # check that the image is conformed, i.e. LIA orientation
+    if not is_conform(orig, vox_size=None, img_size=None, orientation="lia"):
         logger.info("Internally conforming orig to soft-LIA.")
-        orig = conform(orig, vox_size=None, img_size=None, orientation=None)
+        orig = conform(orig, vox_size=None, img_size=None, orientation="lia")
```
dkuegler: We cannot require the orientation to be lia here, that breaks all --keepgeom behavior. What we can do is re-orient to soft-lia
  
```suggestion
    # check that the image is conformed, i.e. LIA orientation
    if not is_conform(orig, vox_size=None, img_size=None, orientation=None):
        logger.info("Internally conforming orig to soft-LIA.")
        orig = conform(orig, vox_size=None, img_size=None, orientation=None)
```
The issue with reorienting to soft-lia is that orig no longer has the same affine (vox2ras) as the underlying file...
So essentially the logger.info string was also wrong in practice.
In reality, I think that we only want to make sure the intensities are robustly scaled.

CorpusCallosum/shape/postprocessing.py
```diff
@@ -430,82 +431,87 @@ def recon_cc_surf_measure(
...
-    # clean up cc mask
-    cc_mask = connect_diagonally_connected_components(cc_mask_slice)
-    # create a CCContour from the cc_mask and transform to RAS coordinates
-    # - R coordinate is stored in _contour.z_position
-    # - AS coordinates are stored in _contour.points
-    _contour = CCContour.from_mask_and_acpc(
-        cc_mask, ac_coords_vox, pc_coords_vox,
-        slice_vox2ras=slice_lia_vox2midslice_ras, contour_smoothing=contour_smoothing,
-    )
+    try:
```
dkuegler: Why was this function changed to return None if it fails, why do we not just use a try block around this "externally", i.e. where it gets called.
This basically means to put a mini function in ~line 191

```python
    def _zip_failed(it_idx: Iterable[int], it_affine: Iterable[AffineMatrix4x4], it_result: Iterable[tuple[CCMeasuresDict, CCContour]]) -> Generator[tuple[int, AffineMatrix4x4, tuple[CCMeasuresDict, CCContour]]]:
        for idx, affine in zip(it_idx, it_affine, strict=True):
            try:
                yield idx, affine, next(it_result)
            except StopIteration:
                # Last Item
                logger.error("Number of items in idx and affine did not match results")
            except Exception as e:
                    logger.error(f"Processing slice {idx} failed: {e.args[0]}")
                    logger.exception(e)
```

This `_zip_failed` can be used instead of the `zip` in line ~191.

The current version cannot raise an exception, so the "Raises" part of the docstring is inconsistent with the implementation.

CorpusCallosum/shape/postprocessing.py
```diff
@@ -430,82 +431,87 @@ def recon_cc_surf_measure(
...
-        "total_area": total_area,
-        "total_perimeter": total_perimeter,
-        "split_contours": split_contours,
-        "subdivision_lines": subdivision_lines,
-        "midline_equidistant": midline_equi,
-        "levelpaths": levelpaths,
-        "slice_index": slice_idx
-    }
-    return measures, _contour
+        measures: CCMeasuresDict = {
```
dkuegler: It seems to me this line is currently not indented correctly (maybe it was also broken before or is not showing correctly in my review view).

dkuegler (commented):
I am also going to bring in a small fix of fastsurfer_cc I did yesterday into this PR.

CorpusCallosum/data/fsaverage_data.json
dkuegler, line 1: Quick question: did we just have bad numbers in this file or what happened here? 10000000000 seems like a debug number ...

FastSurferCNN/utils/common.py
ClePol: feel free to revert then. I would recommend adding a check and error for relative paths instead though

CorpusCallosum/fastsurfer_cc.py
ClePol: We can do it, but before none is returned there should already be an error or warning written to the logfile

CorpusCallosum/data/fsaverage_data.json
ClePol: The file was originally created by me, by reading the header with nibabel and writing it to json. I did not trace back why the number was in there

ClePol:
I think quite a few review comments are refactoring and require changing multiple files and then another full testing. @dkuegler If you are adding another change and doing testing, it would be great if you could also address your comments at the same time

dkuegler:
> I think quite a few review comments are refactoring and require changing multiple files and then another full testing. @dkuegler If you are adding another change and doing testing, it would be great if you could also address your comments at the same time

I am not quite sure that is true... Most of them directly changed that are part of this PR...

The change I added can be seen in my fork in `fix/FastSurferCC/errorcode`. I do not want to mix my changes with yours though. I'd like to keep the commit history clean.
My changes do not have any overlap with yours as far as I can see.

FastSurferCNN/utils/common.py
dkuegler: Well. The intended API is clear: relative paths are relative to the subject's directory. I'll see if I can improve the documentation and error handling...

Issue #790, opened 2026-02-17
Total gray matter estimation from aseg+DKT.stats without surface reconstruction


julii-a:
General Support Request
=======================

Description
-----------
Hello, I have a question about estimating total gray matter volume when using only the FastSurfer segmentation module.

As I understand, **aseg+DKT.stats** is generated from the segmentation module, while **aseg.stats** (including total gray matter) is produced after the surface module. Therefore, the total gray matter value in aseg.stats may be influenced by surface reconstruction (e.g., pial refinement).

If I want to use only segmentation-based volumetric outputs (e.g., aseg+DKT.stats), what is the recommended approach to obtain total gray matter volume?

Specifically:

1. Is it acceptable to derive total gray matter by summing relevant cortical and subcortical regions from aseg+DKT.stats?

In the FreeSurfer documentation, **TotalGray** is described as the sum of *lhCortex + rhCortex + SubCortGray + CerebellumGM*, which includes both surface-based volume calculations and voxel counts. Based on this, I am considering computing total gray matter by summing the corresponding cortical, subcortical, and cerebellar gray matter volumes from aseg+DKT.stats to keep the analysis fully volumetric.

2. Is there a recommended or standard list of regions to include for this purpose when using FastSurfer segmentation outputs?
3. Alternatively, is it still recommended to use the total gray matter value from aseg.stats, even when focusing on segmentation-based volumetric analysis?

I could not find clear guidance in the FreeSurfer documentation, and since outputs like aseg+DKT.stats seem to be specific to FastSurfer, I wanted to ask about the recommended practice.

Thank you for your guidance.

Environment
-----------
FastSurfer Verions: 2.4.2
Installation type: Docker
OS: Windows (WSL2, Ubuntu 22.04)

dkuegler:
Hello @julii-a,


Let me answer your questions real quick:

1. That would be acceptable, and should also be good and reliable. In fact, we have a way to calculate this number for you in FastSurfer, we just do not enable it by default, because it might lead to confusion about why there are two numbers for the same thing (Total Gray Matter).
   The easiest (and consistent way) would be to from `stats/aseg.VINN.stats` find/read the following lines and add the volumes:
   ```
   ...
   # Measure SubCortGray, SubCortGrayVol, Subcortical gray matter volume, 59860.993064, mm^3
   ...
   2   3    734784 245592.265  Left-Cerebral-Cortex              77.6681     7.3268    50.0000   110.0000    60.0000
   ...
[...]

PR #792, opened 2026-02-19
Add missing files to the documentation

dkuegler:
The documentation had a couple of files missing, specifically I became aware of this in #790, when I wanted to point to a function in `brainvolstats`. This PR adds missing python files to the sphinx documentation, fixes their docstrings so the documentation builds and finally adds missing or incomplete docstrings in `brainvolstats`.


m-reuter:
in my actions I moved to python 3.13, also in build matrix I now do 3.10 , 3.11, 3.12 and 3.13 (dropping 3.9). And remove 
architecture: 'x64' as that is too restrictive and collides with Mac architectures .

dkuegler:
Don't really need to test different versions to build the documentation... But I think we might want to move to UV in actions as well, I'll let copilot work on that tomorrow...

PR #799, opened 2026-03-16
Fix: symlink may point to absolute path

dkuegler:
This PR fixes an error in run_fastsurfer.sh, where $subject_dir/stats/aseg+DKT.stats was not pointing to the correct file. aseg+DKT.stats is usually a symlink to aseg+DKT.VINN.stats, but the target was previously an absolute path, which would create invalid paths if created inside a docker or singularity image that mounted $subject_dir somewhere else. Another failure "opportunity" was when files were moved.

Furthermore, the copy option of softlink_or_copy would fail, if a relative path was used, but the current working directory was not the directory of the link.

### DIFF
--- a/recon_surf/functions.sh
+++ b/recon_surf/functions.sh
@@ -225,15 +225,16 @@ function check_allow_root()
 
 function softlink_or_copy()
 {
+  # Creates a symlink at file pointing to target; if that fails, for example because symlinks are not supported by the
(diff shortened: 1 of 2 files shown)


dkuegler:
The doc build error is a problem inherited from pyrr, and this PR does not have any changes to the doc. (can be ignored)

PR #802, opened 2026-04-21
CorpusCallosum registration with neuroreg

ClePol:
This introduces neuroreg as a dependency and replaces FastSurferCC centroid registration with it's modules.


m-reuter:
In case that the number of centroids in fsaverage is changed in neuroreg in the future, it would be good if this caller specifies a list of centroids that should be used for alignment explicitly.

m-reuter:
Superseded by #811

PR #816, opened 2026-05-19
Release docker changes

ClePol:
## Summary

  This PR adds support for building Docker images from the pinned, backend-neutral `requirements.txt` instead of always
  resolving dependencies from `pyproject.toml`.

  It also updates the generated `requirements.txt` from the tested release apptainer image and introduces an error message when users forget to map their host UID/GID into the Docker container instead of a crash.

  ## Changes

  - Add `tools/export_pip-r.sh` support for exporting backend-neutral pinned requirements from Docker or Singularity/
  Apptainer images.
  - Regenerate `requirements.txt` from `fastsurfer-dev-3c52a21.sif`.
  - Add `tools/Docker/build.py --pinned_requirements`.
  - Pass `PINNED_REQUIREMENTS=true` into the Dockerfile and install via `uv pip compile --torch-backend=<device>
  requirements.txt`.
  - Keep CUDA/backend wheel selection separate from the pinned shared dependency set.
  - Include root `requirements.txt` in the Docker build context despite the broader `.dockerignore` requirements pattern.
  - Create a real default `nonroot` user in the Docker image.
  - Reject runs as the default Docker `nonroot` user with a message asking users to pass `-u $(id -u):$(id -g)`.
  - Keep existing root protection and `--allow_root` behavior.

  ## Testing

  - Built Docker with `--pinned_requirements`.
  - Verified image metadata and default user.
  - Verified default unmapped Docker run exits with the new `-u $(id -u):$(id -g)` message.
  - Verified root run exits with the existing root guard.
  - Verified root with `--allow_root` warns and continues.
  - Verified mapped host user run works and reaches the expected license check.
  - Verified `long_fastsurfer.sh` also triggers the new default nonroot guard.
  - Verified Python CUDA stack: `torch 2.7.1+cu118`, `torchvision 0.22.1+cu118`, CUDA `11.8`.
  - Verified `--pinned_requirements --dry_run` emits `PINNED_REQUIREMENTS=true` for another CUDA backend.


tools/export_pip-r.sh
```diff
@@ -19,12 +19,14 @@ function usage()
...
-the FastSurfer repository.
+Will generate a backend-neutral requirements file from docker/singularity images
+to distribute with the FastSurfer repository. Backend-specific PyTorch wheel
+suffixes, PyTorch index URLs, nvidia-* runtime packages, and triton packages are
+intentionally omitted; select those at install/build time with uv --torch-backend.
 
 examples:
 export_pip-r.sh $FASTSURFER_HOME/requirements.txt fastsurfer:dev
-export_pip-r.sh $FASTSURFER_HOME/requirements.cpu.txt fastsurfer:dev
+export_pip-r.sh $FASTSURFER_HOME/requirements.txt fastsurfer.sif
```
m-reuter, line 31: why a .sif ?

tools/export_pip-r.sh
```diff
@@ -39,41 +41,82 @@ fi
...
+} > "$tmp_target"
 
 pip_cmd="python --version && pip list --format=freeze --no-color --disable-pip-version-check --no-input --no-cache-dir"
-if [ "${2/#.sif}" != "$2" ]
+if [[ "$image" == *.sif ]]
 then
   # singularity
-  cmd=("singularity" "exec" "$2" "/bin/bash" -c "$pip_cmd")
-  clean_cmd="singularity exec $2 /bin/bash -c '$pip_cmd'"
+  if command -v singularity > /dev/null
```
m-reuter: I think we should use apptainer first and singularity as fallback. and document this somewhere.

requirements.txt
```diff
@@ -1,120 +1,105 @@
 #
-# This file is autogenerated by kueglerd from fastsurfer:2.4.0-rc1
+# This file is autogenerated by pollakc from /Users/pollakc/Repositories/fastsurfer-dev-3c52a21.sif
```
m-reuter: I don't like local paths in public commits. Remove path print out. same below.

ClePol: changed (didn't forcepush to remove though)

tools/export_pip-r.sh
ClePol: swapped

PR #818, opened 2026-05-21
Surface recon determinism

ClePol:
## Summary

  This PR contains 
1. An improvement of the surface pipeline setting topo fixer to single thread
2. Single threading in spherical project

  ## Testing

End-to-end pipeline run before and after changes on 5 cases.

### DIFF
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -794,7 +794,7 @@ for hemi in lh rh ; do
       echo "echo \"\""
     } | tee -a "$CMDF"
 
-    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag $fsthreads"
+    cmd="env OMP_NUM_THREADS=1 ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS=1 recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag -threads 1 -itkthreads 1"
     RunIt "$cmd" "$LF" "$CMDF"
 
     # fix the surfaces if they are corrupt
--- a/recon_surf/recon-surf.sh
+++ b/recon_surf/recon-surf.sh
@@ -794,7 +794,7 @@ for hemi in lh rh ; do
       echo "echo \"\""
     } | tee -a "$CMDF"
 
-    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag -threads 1 -itkthreads 1"
+    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag $fsthreads"
     RunIt "$cmd" "$LF" "$CMDF"
 
     # fix the surfaces if they are corrupt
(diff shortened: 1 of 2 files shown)


recon_surf/recon-surf.sh
```diff
@@ -794,7 +794,7 @@ for hemi in lh rh ; do
       echo "echo \"\""
     } | tee -a "$CMDF"
 
-    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag $fsthreads"
+    cmd="env OMP_NUM_THREADS=1 ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS=1 recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag -threads 1 -itkthreads 1"
```
m-reuter:
```suggestion
    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag -threads 1 -itkthreads 1"
```
OMP and ITK env not needed as recon-all should set them via the threads flags

m-reuter:
my tests: 
- blas on linux gives slightly different outputs depending on the number of threads, so keeping that fixed to one thread is a good idea (also comes at almost no cost as the mesh is relatively small for the solver). 
- the env var about numexpr is dead weight as numexpr is not a dependency here. The others don't hurt to keep. 
- for the topofixer on a more tricky case I found that 1,1-repeat, 2, 4, 8 and 8-repeat, all give identical results if started from the same initial surface. Speed wise you get 15% improvement going from 1 to 4 threads where it plateaus . So I see no reason to fix it to 1 single thread here. 

I will modify the PR accordingly and then merge it. What then remains is to test if the full surface pipeline is now deterministic.

recon_surf/recon-surf.sh
```diff
@@ -794,7 +794,7 @@ for hemi in lh rh ; do
       echo "echo \"\""
     } | tee -a "$CMDF"
 
-    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag $fsthreads"
+    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag -threads 1 -itkthreads 1"
```
m-reuter:
```suggestion
    cmd="recon-all -subject $subject -hemi $hemi -fix -no-isrunning -umask $(umask) $hiresflag $fsthreads"
```

m-reuter: fixing to 1 thread does not seem to be needed

recon_surf/spherically_project_wrapper.py
```diff
@@ -48,8 +48,20 @@ def setup_options():
...
     opts = setup_options()
 
+    # The spectral projection is sensitive to tiny threaded BLAS/eigensolver
+    # differences, and those can be amplified by later topology correction.
+    for var in (
+        "OMP_NUM_THREADS",
+        "OPENBLAS_NUM_THREADS",
+        "MKL_NUM_THREADS",
+        "NUMEXPR_NUM_THREADS",
+        "VECLIB_MAXIMUM_THREADS",
```
m-reuter:
```suggestion
        "VECLIB_MAXIMUM_THREADS",
```

PR #820, opened 2026-05-22
Speed up FastSurfer processing

ClePol:
This is an AI driven change of the FastSurfer pipeline, aimed to speedup the segmentation and surface processing components. Current acceleration is around 25-35% depending on the number of threads available. The speedup of single changes - in an uncontrolled test environment with --threads 8 - are included in most commits and as detailed commit messages.

Before merging this PR the maintainers of the specific modules should review and edit the changes if needed.

I also recommand running the full validation prior to merging as this PR touches many core components - initial tests show no regression.

Note that [c81f0c7](https://github.com/Deep-MI/FastSurfer/pull/820/commits/c81f0c7456c040ee00f8a561fc9d3cc0575ba3b6) disables writing the usually unused T1.mgz and allows enabling it instead of disabling it.

### DIFF
--- a/CerebNet/inference.py
+++ b/CerebNet/inference.py
@@ -34,6 +34,7 @@ from FastSurferCNN.utils.arg_types import ImageSizeOption, OrientationType
 from FastSurferCNN.utils.common import SubjectDirectory, SubjectList, find_device
 from FastSurferCNN.utils.mapper import JsonColorLookupTable, Mapper, TSVLookupTable
 from FastSurferCNN.utils.parallel import SerialExecutor, get_num_threads
+from FastSurferCNN.utils.torchscript import cpu_torch_threads
 
 if TYPE_CHECKING:
     import yacs.config
@@ -92,7 +93,6 @@ class Inference:
         self._threads = None
         self.threads = threads
         _threads = get_num_threads() if self._threads is None else self._threads
-        torch.set_num_threads(_threads)
         self.pool = ThreadPoolExecutor(self._threads) if async_io else SerialExecutor()
         self.cfg = cfg
         self._async_io = async_io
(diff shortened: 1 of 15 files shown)


m-reuter:
There are probably some things in here worth considering, but:
- most is parallelisation of a few blocks (each saving very little, like 30s) and adding a lot of additional code and complexity. Probably not worth the added administrative and maintenance overhead. 
- a lot is just faster because it calls some freesurfer commands directly instead of piggy-backing on recon-all. The reason that the slower recon-all is used, is to inherit future updates in these blocks directly from freesurfer.
- some changes modify outputs, but the AI does not care and looks only at the standard outputs (it interprets the files as intermediate files but they are final output, like the different surface curvature files).
- it skips things that are not relevant for the FastSurfer pipeline (like the T1.mgz creation) but are necessary for some downstream tools, like fmri-prep. They had been added for that purpose only. (If this has changed in the meantime in fMRI-prep, we can of course drop it). 

So most of it is overlap (parallelisation) of components, including bigger modules like neural networks. For that of course the GPU needs to handle that. We should consider moving the pipeline to snakemake instead of hacking parallelism into it like here. 

Still worth to browse through these, as some (at least one , the aparc smoothing filter for example) are worth adding, or point to potential places to improve speed in the future. So this PR will never get merged (as is), but can remain open as a guide for place to look for future improvements.

dkuegler:
Generally, a lot of the parallelism circumvents the intended functionality of the `--threads*` flags.

As for the changes to bash scripts, I have tried to avoid these changes in the past, because it makes the (already complex) scripts more complex and I would want to switch over the code from bash to python no matter what as well.

The torchscript extensions might be interesting to consider.

Other scripts I did not take a look at.

PR #821, opened 2026-05-22
Fix wrong conform arguments for T2 and numpy2 related change

ClePol:
This should fix two bugs in stable where FastSurfer fails when given a T2. Conform.py seems has changed its arguments, but they were not adjusted for the T2 processing. 

Creating this as a draft to be edited by the module maintainers. I only ran testing for one case.

### DIFF
--- a/run_fastsurfer.sh
+++ b/run_fastsurfer.sh
@@ -621,7 +621,6 @@ then
 fi
 
 check_create_subjects_dir_properties "$sd"
-export SUBJECTS_DIR="$sd"
 
 if [[ -z "$subject" ]]
 then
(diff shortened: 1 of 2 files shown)


run_fastsurfer.sh
```diff
@@ -621,6 +621,7 @@ then
 fi
 
 check_create_subjects_dir_properties "$sd"
+export SUBJECTS_DIR="$sd"
```
```suggestion
```

m-reuter:
Setting $SUBJECTS_DIR is not needed for registration (mri_coreg or mri_robust_register, nor mri_vol2vol). The other bug is related to our upgrade to numpy2 where ndarray.newbyteorder() was removed.

PR #828, opened 2026-05-26
Use neuroreg for registration, mapping, conversion in several places (HypVINN, longitudinal primarily)

m-reuter:
In order to remove the FreeSurfer dependencies in the seg_only path, we replace `mri_coreg`, `mri_robust_register`, `mri_robust_template`, `mri_vol2vol`, `mri_mask`, `mri_binarize`, `mri_info`) calls with `neuroreg`.

- this speeds up processing
- with `--keep-dtype` we can keep the input data type to replicate FreeSrufer's behaviour
- similary we use cubic interpolation to map the movable images after registration as default
- in hypvinn module we drop the option for `mri_robust_register --cost nmi` as we now use neuroreg.coreg which also replaces `mri_coreg` at the same time (cross modal registration based on NMI and Powell optimiser). 

In the future, we can generate quick WM surfaces based on the segmentaiton and then can explore if `neuroreg bbreg` adds any value here.

### DIFF
--- a/recon_surf/long_prepare_template.sh
+++ b/recon_surf/long_prepare_template.sh
@@ -226,6 +226,8 @@ fi
 # check that SUBJECTS_DIR exists
 check_create_subjects_dir_properties "$SUBJECTS_DIR"
 
+auto_detect_fs_license "the longitudinal template preparation" || exit $?
+
 ################################## SETUP and LOGFILE ##############################
 
 
@@ -455,5 +457,4 @@ do
   cmd="mri_convert -at ${ltaXforms[$i]} -rt $interpol $mdir/cross_input${extension} $mdir/long_conform${extension}"
   RunIt "$cmd" "$LF"
 done
-
 
--- a/recon_surf/long_prepare_template.sh
+++ b/recon_surf/long_prepare_template.sh
@@ -39,7 +39,7 @@
 #   - Add flag for adding a new time point to an existing base/template.
 #
 #  FreeSurfer requirements: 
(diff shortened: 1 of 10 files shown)


recon_surf/long_prepare_template.sh
```diff
@@ -401,38 +405,64 @@ done
...
+  # create the 'mean/median' input (orig) volume by reusing the transforms above (no
+  # registration, --noit); the per-tp LTAs carry the base geometry so orig.mgz lands on
+  # the same grid as base_brainmask:
+  cmd="$python -m neuroreg.cli.multireg --mov ${subjInVols[*]}"
   cmd="$cmd --average ${robust_template_avg_arg}"
   cmd="$cmd --ixforms ${ltaXforms[*]}"
   cmd="$cmd --noit"
   t1=${SUBJECTS_DIR}/$tid/mri/orig.mgz
-  cmd="$cmd --template $t1"
+  cmd="$cmd --template $t1 $device_opt"
```
m-reuter:
```suggestion
  cmd="$cmd --template $t1"
```

m-reuter: device not needed here, as we are not registering, only mapping, which happens on the CPU.

requirements.txt
```diff
@@ -60,7 +60,7 @@ multipledispatch==1.0.0
 narwhals==2.21.0
 networkx==3.6.1
 neurolit==0.6.1
-neuroreg==0.6.2
+neuroreg==0.10.1
```
m-reuter:
```suggestion
neuroreg==0.10.4
```

PR #841, opened 2026-07-09
Add BIDS entrypoint, smoke tests, and routing fixes

karl-koschutnig:
## Description

This PR adds initial BIDS support to FastSurfer via a new run_fastsurfer_bids.py entrypoint.

## Scope

- discover subjects/sessions from a BIDS dataset using pybids
- route single-session cases to brun_fastsurfer.sh
- route multi-session cases to long_fastsurfer.sh
- support participant/session filtering
- write a minimal derivatives dataset_description.json

## Tests / developer workflow

- added unit coverage for BIDS discovery/routing
- added an OpenNeuro smoke test based on ds004937
  - one single-session subject
  - one multi-session subject
- added a small developer wrapper to run the BIDS smoke checks quickly

## Routing fix

During testing, --seg_only exposed an invalid path where multi-session BIDS subjects were still being sent to long_fastsurfer.sh, which does not support --seg_only / --surf_only.

This PR fixes that by forcing cross-sectional routing when those passthrough flags are present.

## Additional fix included

While validating cross-sectional runs, the Corpus Callosum pipeline failed under NumPy 2.x because it used np.cross() on 2D vectors. This PR replaces those calls with an explicit signed 2D cross-product and adds a regression test.

## Validated

- targeted pytest coverage for BIDS logic
- OpenNeuro dry-run smoke test
- --seg_only cross-sectional routing
- mixed dry-run routing (brun + long)
- CC NumPy 2.x regression test

## Out of scope / missing

- no BIDS-aware srun_fastsurfer.sh integration yet
- SLURM support would need separate orchestration for cross-sectional vs longitudinal BIDS cases

### DIFF
--- a/CorpusCallosum/shape/subsegment_contour.py
+++ b/CorpusCallosum/shape/subsegment_contour.py
@@ -880,8 +880,8 @@ def transform_to_acpc_standard(
     norms_product = np.linalg.norm(ac_pc_vec) * np.linalg.norm(posterior_vector)
     theta = np.arccos(dot_product / norms_product)
 
-    # Determine the sign of the angle using cross product
-    cross_product = np.cross(ac_pc_vec, posterior_vector)
+    # Determine the sign of the angle using the signed z-component of the 2D cross product.
+    cross_product = ac_pc_vec[0] * posterior_vector[1] - ac_pc_vec[1] * posterior_vector[0]
     if cross_product < 0:
         theta = -theta
 
--- a/CorpusCallosum/utils/mapping_helpers.py
+++ b/CorpusCallosum/utils/mapping_helpers.py
@@ -84,8 +84,8 @@ def correct_nodding(ac_pt: Vector2d, pc_pt: Vector2d) -> AffineMatrix3x3:
     norms_product = np.linalg.norm(ac_pc_vec) * np.linalg.norm(posterior_vector)
     theta = np.arccos(dot_product / norms_product)
 
-    # Determine the sign of the angle using cross product
-    cross_product = np.cross(ac_pc_vec, posterior_vector)
+    # Determine the sign of the angle using the signed z-component of the 2D cross product.
+    cross_product = ac_pc_vec[0] * posterior_vector[1] - ac_pc_vec[1] * posterior_vector[0]
     if cross_product < 0:
         theta = -theta
 
--- a/FastSurferCNN/utils/bids.py
+++ b/FastSurferCNN/utils/bids.py
@@ -1,0 +1,160 @@
+# Copyright 2026 Image Analysis Lab, German Center for Neurodegenerative Diseases (DZNE), Bonn
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Discovery and grouping of subjects/sessions from a BIDS dataset for run_fastsurfer_bids.py."""
+
+import json
+import logging
+from dataclasses import dataclass, field
+from pathlib import Path
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class BidsSession:
+    """A single anat session found for a subject (may be the subject's only session)."""
+
+    session_id: str | None
+    t1w: Path
+    t2w: Path | None = None
+
+
+@dataclass
+class BidsSubject:
+    """A subject with one or more sessions, plus its processing mode."""
+
+    subject_id: str
+    sessions: list[BidsSession] = field(default_factory=list)
+
+    @property
+    def is_longitudinal(self) -> bool:
+        """Return True if this subject has more than one session with a T1w image."""
+        return len(self.sessions) > 1
+
+    def output_id(self, session: BidsSession) -> str:
+        """Return the FastSurfer subject/timepoint id for a given session of this subject."""
+        if session.session_id is None:
+            return self.subject_id
+        return f"{self.subject_id}_ses-{session.session_id}"
+
+
+def _require_pybids():
+    try:
+        import bids as pybids
+    except ImportError as e:
+        raise ImportError(
+            "The 'bids' (pybids) package is required for BIDS support but is not installed. "
+            "Install it with `pip install fastsurfer[bids]` or `pip install pybids`."
+        ) from e
+    return pybids
+
+
+def find_subjects(
+    bids_dir: Path,
+    participant_labels: list[str] | None = None,
+    session_labels: list[str] | None = None,
+    validate: bool = True,
+) -> list[BidsSubject]:
+    """
+    Discover subjects, sessions, and their T1w/T2w images in a BIDS dataset.
+
+    Parameters
+    ----------
+    bids_dir : Path
+        Path to the root of a BIDS-valid dataset.
+    participant_labels : list[str], optional
+        If given, only include these subject labels (without the 'sub-' prefix).
+    session_labels : list[str], optional
+        If given, only include these session labels (without the 'ses-' prefix).
+    validate : bool, default=True
+        Whether pybids should validate the dataset against the BIDS spec.
+
+    Returns
+    -------
+    list[BidsSubject]
+        Subjects found, each with the sessions that have a usable T1w image. Subjects
+        with no T1w anywhere in the dataset are omitted (with a warning).
+    """
+    pybids = _require_pybids()
+    layout = pybids.BIDSLayout(str(bids_dir), validate=validate)
+
+    subject_ids = layout.get_subjects()
+    if participant_labels:
+        wanted = {label.removeprefix("sub-") for label in participant_labels}
+        missing = wanted - set(subject_ids)
+        if missing:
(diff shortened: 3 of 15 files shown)


m-reuter:
Hi , thanks, we will take a look. One thing up front, fixes should not be mixed into feature PRs . Can you do a separate PR for the CC numpy 2 fix (and @ClePol  can check that). After merging it, we can rebase this one.

karl-koschutnig:
Thanks, Martin, for your prompt response.
I’ve created a separate PR.
Best,
Karl

m-reuter:
Thanks for this, and sorry for the slow reply. We do want a BIDS entry point, and two of your
choices are right: the BIDS-App positional contract cannot be bolted onto `brun_fastsurfer.sh` or
`srun_fastsurfer.sh` without wrecking them, so a separate script is the correct shape, and
delegating to the existing scripts rather than reimplementing the pipeline is what I would have
asked for.

It cannot go in as it stands, and probably getting it there is a fair bit of work.

**1. `--seg_only` / `--surf_only` should not drop a subject to cross-sectional.** The longitudinal
pipeline supports both of those functionalities, it just spells them as stages:
`--stage prepare --stage template_seg --stage long_seg` for segmentation only, and
`--stage template_surf --stage long_surf` for surfaces only. #906 added the message that tells the
user exactly this. So what `long_fastsurfer.sh` does not accept is the flag names, not the
operation. Falling back to cross-sectional instead swaps the scientific method rather than the
spelling: timepoints stop being conditioned on the person-template, so the aseg differs, and
nothing says so. Mapping the two flags onto those stages is the fix.

**2. Longitudinal versus cross-sectional, genuinely open.** One part is not open:
In a longitudinal study, single-timepoint participants must also go through the longitudinal stream,
otherwise their results are not comparable to the multi-timepoint ones. Deciding per subject from
the session count does the opposite, sending exactly those cases to cross-sectional.
What we have not settled is how the entry point should let you say it: my instinct is a flag for the
whole run rather than an inference per subject. 

**3. pybids is heavy for what it is used.** We need subject, session and T1w/T2w, which is a glob and
a filename split. pybids lands in the Docker image, the macOS bundle and our dependency baseline.
Best implemented without such heavy dependencies if that is at all possible.

Smaller things:

- Important: Output naming, `sub-01_ses-1` flat or nested under `output_dir`, is the decision hardest
  to reverse later, since it is what downstream BIDS tooling consumes. Worth stating in `BIDS.md`.
- `srun_fastsurfer.sh` is not a target. Cluster users are most of the people with BIDS datasets big
  enough to want this. It should probably be added.
- The OpenNeuro scripts are fine as they are, dry-run by default and ~60 MB. But
  `test/lint/test_every_test_dir_runs.py` requires every test directory to be reached by a
  workflow, so `test/integration/` forces the question of what runs in CI. Our preference: wire in
  the dry-run discovery and routing check only, and keep this out of pipelinetest, which we want to
  stay narrow for now.
- `--dry_run` is spelled `--dry` in `srun_fastsurfer.sh`, probably best to align.
- Worth double-checking: `force_long` looks assigned but unused in the routing expression, which
  would make `--longitudinal` a no-op for single-session subjects.

Rebase: the two `CorpusCallosum` hunks are already on `dev` via #872, and `test/quicktest/` is gone,
tests were regrouped into `test/lint`, `shell`, `image`, `utils` and `pipelinetest`, with a lint
requiring every test directory to be reached by a workflow.

If you do want to continue, I would split it: discovery plus the BIDS-App contract, cross-sectional
only, without pybids, has no open design questions and is useful on its own. Longitudinal grouping
then gets its own PR where point 2 can be settled properly.

karl-koschutnig:
Thanks for the detailed read — the split you suggested is the right call, and I've done it that way.

**PR A is #909: discovery plus the BIDS-App contract, cross-sectional only.** It is a fresh branch
off current `dev` rather than a force-push here, so this PR is superseded and I am closing it — the
two `CorpusCallosum` hunks are gone with #872 and `test/quicktest/` with the regrouping.

**3. pybids is gone.** Discovery is the standard library: subject, session and T1w/T2w are what the
BIDS layout spells out in the paths, so `sub-*/[ses-*/]anat/*_T1w.nii*` is a glob and a suffix check.
No `[bids]` extra, so nothing changes in the Docker image, the macOS bundle or the dependency
baseline. `--skip_bids_validator` is still there for the BIDS-App contract; validation shells out to
the `bids-validator` tool where the user has it and says so where they don't, rather than pulling the
specification in as a dependency.

**1 and 2 do not arise here.** With every session processed as its own cross-sectional case there is
no per-subject inference and no fallback, so nothing swaps the method silently. `--seg_only` passes
straight through to `brun_fastsurfer.sh` as any other option does. You were also right about
`force_long` — it was assigned and then overridden by the `elif` chain below it, making `--longitudinal`
a no-op for single-session subjects. That code is gone rather than fixed.

**Output naming is stated in `BIDS.md`,** with a directory tree and the reasoning: `output_dir` is
`SUBJECTS_DIR` and each session is a flat `sub-<label>_ses-<label>` directory in it, not nested. That
is what FreeSurfer tooling expects of a `SUBJECTS_DIR`, so downstream commands work on the output
directory unchanged, and it is the same naming the longitudinal pipeline uses for its timepoints.

**`srun_fastsurfer.sh` is now a target.** `--slurm` swaps the script and nothing else — it already
parses the identical `sid=path --extra args` subject list, so there was nothing to translate.

**`--dry`,** with `--dry_run` still accepted, matching how `srun_fastsurfer.sh` spells it.

**On CI, I went a different way than suggested and would rather flag it than have you find it.**
There is no `test/integration/` pytest directory and no workflow change at all. The dry-run discovery
and routing check lives in `test/utils/test_bids.py`, which the existing unit test workflow already
runs, so the CI diff is zero lines and `pipelinetest` stays as narrow as you want it. Routing is
checked through `--dry`, so it needs no data, no network and no FreeSurfer.

`test/integration/openneuro_check.sh` is in #909 but is not a CI test and no workflow names it —
the directory holds no `test_*.py`, so `test_every_test_dir_runs.py` is satisfied without a download
in CI. It fetches two sessions of a randomly drawn subject from ds004937 (public, no login, ~12 MB a
session), runs them, and checks the outputs against `test/pipelinetest/data/expected-files.yaml`
rather than a list written again by hand. Happy to drop it if you would rather the repository did not
carry it.


**One thing I did not decide, because it is yours to make:** the Docker `ENTRYPOINT` is still
`run_fastsurfer.sh`, so `docker run <image> /bids /out participant` is not a BIDS-App invocation yet.
The script is in the image and reachable with `--entrypoint`. Changing the default is user-facing and
I did not want to assume it — there is a reasonable argument for the BIDS-App form and a reasonable
argument against.

---

**PR B: longitudinal. Here is a concrete proposal for point 2, rather than more code to review.**

`--longitudinal` as a flag for the whole run, as your instinct said. Every selected subject goes
through `long_fastsurfer.sh`, **single-session subjects included** — `long_fastsurfer.sh:344` requires
`-lt 1` rather than `-lt 2`, so a one-timepoint template is a valid call, and that is what keeps those
participants comparable to the multi-timepoint ones. Nothing is inferred from the session count
anywhere. Without the flag, PR A's behaviour is unchanged.

Per subject, sessions sorted by label:

```
long_fastsurfer.sh --tid sub-01 --tpids sub-01_ses-1 sub-01_ses-2 \
                   --t1s <abs ses-1 T1w> <abs ses-2 T1w> --sd <output_dir> [passthrough]
```

**`--seg_only` and `--surf_only` are translated, not refused** — your point 1. Under `--longitudinal`
they are intercepted and become `--stage prepare --stage template_seg --stage long_seg` and
`--stage template_surf --stage long_surf`. The user writes the same flag in either stream; only the
spelling handed to the script changes. Both at once, or either alongside an explicit `--stage`, is an
error rather than a guess.

**Two refusals fall out of the mapping.** A dataset with no `ses-` level cannot be run with
`--longitudinal`: `--tid` and `--tpids` would both be `sub-01`, the same directory in two roles. And
T2w is ignored with a warning naming the subjects, since `long_fastsurfer.sh:222` rejects `--t2`, so
HypVINN is unavailable in this stream.

**A guard against the two streams overwriting each other.** A longitudinal timepoint directory and a
cross-sectional one have the same name and hold different science, so the derivatives
`dataset_description.json` records which stream produced the directory, and a run whose method
disagrees with what is recorded is refused. This is the same class of problem as point 1, one level up.

Subjects run sequentially, one `long_fastsurfer.sh` call each, since there is no `--subject_list`
there; `--parallel` still works within a subject. A subject that fails is recorded and the rest still
run.

---

**`--slurm` with `--longitudinal` would be an error in that PR, and I want to explain why rather than
just leave the gap.**

`srun_fastsurfer.sh` has no longitudinal mode, and what it would do instead is process every timepoint
cross-sectionally — point 1 coming back through a different door. So it has to refuse.

But it is tractable, and the shape looks small enough to be worth saying out loud. The five stages'
dependencies are all within one subject and `long_fastsurfer.sh` already sequences them, so there is
no cross-subject dependency graph to build — subjects are independent jobs. The only real constraint
is that the stages split by hardware, `template_seg`/`long_seg` being `--seg_only` work and
`template_surf`/`long_surf` being `--surf_only` work, which is exactly the split
`srun_fastsurfer.sh` already has partitions, memory and time limits for. Two jobs per subject with one
`afterok` edge covers it:

```
job A (seg partition):   --stage prepare --stage template_seg --stage long_seg
job B (surf partition):  --stage template_surf --stage long_surf   --dependency=afterok:<A>
```

That belongs in `srun_fastsurfer.sh` as a `--longitudinal` mode, not in the BIDS entry point — the
partition handling, the singularity wrapping, the work-directory staging and the cleanup all already
live there, and `run_fastsurfer_bids.py` should delegate to it the same way it delegates the
cross-sectional case today. I am happy to do that work as a separate PR if you want it; I did not want
to build a second slurm driver inside a BIDS entry point without asking.

Let me know on point 2 and I will open PR B accordingly.

karl-koschutnig:
Superseded by #909.

PR #850, opened 2026-08-04
Modify fov computation, particullary the decimal precision on the voxel size

santiestrada32:
the surface pipeline fails during the conform stage when processing orig.mgz. It seems to be related to the FOV calculation and the decimal precision of the voxel size
old recon-surf
[recon-surf.log](https://github.com/user-attachments/files/30697366/recon-surf.log)
after fixing decimal points
[recon-surf.log](https://github.com/user-attachments/files/30697400/recon-surf.log)

### DIFF
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -1603,8 +1603,17 @@ def conformed_vox_img_size(
                 # correct sizes for changing voxel size (if voxel size is changing)
                 # compute field of view dimensions in mm (in native orientation)
                 fov = np.array(np.round(img.header.get_zooms()[:3], decimals=decimals)) * target_img_size
-                # compute number of voxels needed to cover field of view
-                target_img_size = np.ceil((fov / target_vox_size * 10000).astype(int).astype(float) / 10000).astype(int)
+                # number of voxels needed to cover the field of view
+                n_vox = fov / target_vox_size
+                # n_vox is integer when the fov is a multiple of the voxel size, but floating-point
+                # error makes it only approximately so. Storing the zoom as float32 leaves a relative
+                # residual of ~1e-8 that scales with the voxel count, i.e. machine precision rather
+                # than a voxel-size tolerance. Snap counts that are integer within that relative error
+                # and round up only genuine partial voxels.
+                rounded = np.rint(n_vox)
+                target_img_size = np.where(
+                    np.isclose(n_vox, rounded, rtol=1e-6, atol=0.0), rounded, np.ceil(n_vox)
+                ).astype(int)
         # use cube (same size in all directions) with MAX_DIMENSION in each direction as minimum
         if _img_size == "auto":
             target_img_size = np.full_like(np.maximum(MAX_DIMENSION, target_img_size), np.amax(target_img_size))
--- a/FastSurferCNN/data_loader/conform.py
+++ b/FastSurferCNN/data_loader/conform.py
@@ -1566,8 +1566,11 @@ def conformed_vox_img_size(
     target_img_size: IntVector3d | None
     MAX_VOX_SIZE = 1.0
     MAX_DIMENSION = 256
-    # number of decimals to round voxel sizes to, so that vox_eps-sized float noise does not affect results
-    decimals = int(np.ceil(-np.log10(vox_eps)))
+    if vox_eps <= 0:
+        raise ValueError(f"vox_eps must be > 0, got {vox_eps}.")
+    # number of decimals to round voxel sizes to, so that vox_eps-sized float noise does not affect
+    # results; clamp at 0 so that vox_eps > 1 does not round voxel sizes to tens of mm
+    decimals = max(0, int(np.ceil(-np.log10(vox_eps))))
     # this is similar to mri_convert --conform_min, note, vox_size == 'auto' is extra, but not covered by VoxSizeOption
     if isinstance(vox_size, str) and (vox_size := cast(VoxSizeOption, vox_size.lower())) in ["min", "auto"]:
         # find minimal voxel side length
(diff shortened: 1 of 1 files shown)


dkuegler (approved):
Looks fine to me. I guess it also does not matter if you round first and then multiply with img_size.

m-reuter:
I wonder why there is a discrepancy between our own conforming step and the conforming check later. Both should be based on the same code and therefore give the same result ?

m-reuter:
## Why this happens
` conform()`  and ` is_conform()`  disagreed on the target image size for this input because  `conformed_vox_img_size() ` computes the target voxel count from the measured voxel size in the header (` fov = zooms * shape ,  n_vox = fov / target_vox_size `), and headers routinely carry tiny float noise — for example ` 0.8000005`  instead of  `0.8` . Tha small difference per voxel, which accumulates over the image extent, is enough to push  `n_vox`  just over an integer boundary ( `320.0002 `) and make  `ceil()`  round up to 321 instead of 320.
The reason this doesn't show up during conforming itself but does show up on the second call (the surface-stage  is_conform  check) is geometry-dependent: on the original cuboid input, the noisy axis isn't the largest dimension, so  `img_size="auto"`  picking the max across axes masks the round-up. Once the image has been conformed to a `320³` cube, the same noisy axis is now full-length and becomes the max, so its round-up is no longer hidden and we get a different answer.

I considered forcing the header after conforming to be exactly .8 on all axes, but that would overwrite the real sizes and loose that information. Headers should stay accurate to what was actually done to the data. 

The fix in this PR rounds voxel sizes to the same precision ( `vox_eps` , already the tolerance the voxel-size check itself uses) before computing the FOV, so the round-up no longer happens in the first place. This keeps ` conformed_vox_img_size()`  idempotent: the size and voxel-size checks now agree with each other on any input, not just this one, without writing any assumption into the image header.

I will add two changes:
-  re-using decimals and thus also fixing the line-too-long error in ruff CI code-style 
- guard the voxel-count-rounding against float error with a relative tolerance (instead of the existing fixed-tolerance guard with `*1000/int/1000`.

sRassmann:
I think the issue still persists in some cases: 
When processing 1mm images that happen to have noise in the header s.t. the voxel size becomes 1 + eps, the checks still fails:

```
python3 -s /fastsurfer/FastSurferCNN/data_loader/conform.py -i /output/presurg_00164/mri/orig.mgz --check_only --vox_size min --verbose
@#@FSLOADPRE 2026:08:11:19:19:35 FastSurferCNN/data_loader/conform.py N 6 L 56.61 71.99 56.41
[INFO: conform.py: 1712]: Image Conform Parameters:
[INFO: conform.py: 1712]: 
[INFO: conform.py: 1712]: - verbosity: True
[INFO: conform.py: 1712]: - input volume: /output/presurg_00164/mri/orig.mgz
[INFO: conform.py: 1712]: - check only: True
[INFO: conform.py: 1712]: - dtype: uint8
[INFO: conform.py: 1712]: - voxel size: min
[INFO: conform.py: 1712]: - round voxel size to 1mm if > threshold: None
[INFO: conform.py: 1712]: - image size: auto
[INFO: conform.py: 1712]: - affine orientation: lia
[INFO: conform.py: 1712]: - log: stdout only
[INFO: conform.py: 1917]: Reading input: /output/presurg_00164/mri/orig.mgz ...
[INFO: conform.py: 1450]: A conformed image must satisfy the following criteria:
[INFO: conform.py: 1454]:  - Number of Dimensions 3        : GOOD image ndim 3
[INFO: conform.py: 1454]:  - Voxel Size min=1.0x1.0x1.0    : GOOD image 1.0000001x1.0x1.0000001
[INFO: conform.py: 1454]:  - Dimensions auto=256x256x256   : GOOD image dimensions (np.int32(256), np.int32(256), np.int32(256))
[INFO: conform.py: 1454]:  - Orientation LIA               : GOOD affine=[[-1. -0. 0.] [ 0. 0. 1.] [-0. -1. 0.]] => LIA
[INFO: conform.py: 1454]:  - Dtype uint8                   : GOOD dtype uint8
[INFO: conform.py: 1952]: Input /output/presurg_00164/mri/orig.mgz is already conformed! Exiting.

@#@FSTIME  2026:08:11:19:19:35 FastSurferCNN/data_loader/conform.py N 6 e 2.15 S 0.24 U 2.03 P 105% M 405732 F 2 R 76721 W 0 c 16 w 2453 I 10514 O 0L 56.61 71.99 56.41
@#@FSLOADPOST 2026:08:11:19:19:37 FastSurferCNN/data_loader/conform.py N 6
python3 -s /fastsurfer/FastSurferCNN/data_loader/conform.py -i /output/presurg_00164/mri/aparc.DKTatlas+aseg.deep.mgz --check_only --vox_size 1.0000001 --dtype any --verbose
@#@FSLOADPRE 2026:08:11:19:19:38 FastSurferCNN/data_loader/conform.py N 8 L 54.31 71.26 56.26
usage: 
Script to conform an MRI brain image to UCHAR, RAS orientation, 
and 1mm or minimal isotropic voxels

USAGE:
conform.py  -i <input> -o <output> <options>
OR
conform.py  -i <input> --check_only <options>
Dependencies:
    Python 3.10+
    Numpy
    https://www.numpy.org
    Nibabel to read and write FreeSurfer data
    https://nipy.org/nibabel/
Original Author: Martin Reuter
Modified by: David Kügler
Date: May-12-2025
conform.py: error: argument --vox_size: invalid vox_size value: '1.0000001'
Command exited with non-zero status 2
@#@FSTIME  2026:08:11:19:19:38 FastSurferCNN/data_loader/conform.py N 8 e 1.88 S 0.19 U 1.98 P 115% M 402920 F 0 R 76051 W 0 c 11 w 11 I 0 O 0L 54.31 71.26 56.26
@#@FSLOADPOST 2026:08:11:19:19:40 FastSurferCNN/data_loader/conform.py N 8
```

I think the problem is that `--vox_size` has to strictly be <= 1.0, i.e. either the parsed voxel size needs to be rounded and forwarded to the second check or choices for `--vox_size`  also need a tolerance.

m-reuter:
@sRassmann see #854 , should fix this.

PR #852, opened 2026-08-11
Tolerate voxel size rounding near one

ClePol:
This PR adresses a bug for recon_surf with voxel sizes slightly above 1.0 (e.g. 1.0000001). 
```bash
conform.py -i "$t1" --check_only --vox_size min --verbose
```
suceeds, but then this turns into 
```bash
conform.py \
    -i "$asegdkt_segfile" \
    --check_only \
    --vox_size 1.0000001 \
    --dtype any \
    --verbose
```
which fails with
```
  conform.py: error: argument --vox_size:
  invalid vox_size value: '1.0000001'
```

## Edits

Normalized numeric `vox_size` values within an absolute tolerance of `1e-6` from 1 mm before applying the existing strict validator. Values materially above 1 mm remain invalid, and the shared `float_gt_zero_and_le_one()` validator remains unchanged. Existing `none` and `infinity` aliases are handled before numeric conversion.

### DIFF
--- a/FastSurferCNN/utils/arg_types.py
+++ b/FastSurferCNN/utils/arg_types.py
@@ -83,31 +83,34 @@ def string_to_bool(a: str) -> bool:
 
 def vox_size(a: str | float | None) -> VoxSizeOption:
     """
-    Convert the vox_size argument to 'min' or a valid voxel size.
+    Convert the vox_size argument to a valid voxel-size option.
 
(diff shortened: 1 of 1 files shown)


m-reuter:
closing as duplicate of #854 which is more complete also treating the python path not only argparse and adds some testing.

PR #853, opened 2026-08-12
Add --extra_slurm_options to srun_fastsurfer.sh for sbatch scheduling options

### LINKED ISSUE #586: Added Support for --reservation Flag for srun_fastsurfer.sh
## Feature Description

I am proposing adding support for a `--reservation` flag to the `srun_fastsurfer.sh` script. This would allow users to specify a reservation when submitting jobs to SLURM clusters using reserved GPU nodes.

## Motivation

The FastSurfer script does not currently have an option to handle SLURM jobs that need to run on reserved GPU nodes. Without a `--reservation` flag, users have to manually modify the script, as I did, to submit jobs using a reservation. Adding this feature would streamline the process for users who need to specify a reservation on clusters with dedicated resources.

Although not everyone may need reserved nodes, I believe this would streamline the workflow for those who do.

## Additional Context

I have already implemented this change in my local version of the `srun_fastsurfer.sh` script. Users would define the reservation as:
`--reservation <reservation_name>
`

The changes I made locally:
Define the reservation variable at the beginning of the script:
`reservation=""`
Add a new case for the --reservation flag in the argument parser:
```
--reservation)
  reservation="$1"
  shift
  ;;
```
Update the SLURM job submission section to include the reservation flag if it is defined:
```
if [[ -n "$reservation" ]]; then
  seg_slurm_sched=(--reservation="$reservation" "${seg_slurm_sched[@]}")
fi

```

The changes I made are very simple and short, but I hope they help to make FastSurfer even better for users with similar needs.
I’ll leave it up to you to decide whether this would be useful to add to the official script.

eastagiletracker:
This PR proposes `--extra_slurm_options`, `--extra_slurm_options_seg` and `--extra_slurm_options_surf` for `srun_fastsurfer.sh`, so scheduling options such as `--reservation` actually reach the `sbatch` calls of the segmentation and surface jobs (fixes #586). 

## What was wrong

`srun_fastsurfer.sh` forwards every option it does not recognize to FastSurfer itself (`POSITIONAL_FASTSURFER`), so a scheduling option is not rejected — it is silently moved into the FastSurfer call inside the container. On `dev` at `ef190b9`, asking for a reserved node produces jobs that are scheduled without the reservation and then abort inside the container:

```console
$ ./srun_fastsurfer.sh --data data --sd out --work work --pattern '*.nii.gz' \
    --fs_license license.txt --singularity_image fastsurfer.sif \
    --reservation gpu_nodes --dry
sbatch --parsable --mem=7G --cpus-per-task=16 -J FastSurfer-Seg- --time=20 -o WORK/logs/seg_%A_%a.log WORK/scripts/slurm_cmd_seg.sh --gpus-per-task=1
...
  /data/scripts/brun_fastsurfer.sh  --subject_list /data/scripts/subject_list --statusfile /data/scripts/subject_success --sd /data/cases --threads 16 --seg_only --reservation gpu_nodes --fs_license /data/scripts/.fs_license
```

The `sbatch` line carries no reservation, while `--reservation gpu_nodes` ends up on the `brun_fastsurfer.sh` command line, where `run_fastsurfer.sh` stops with `ERROR: Flag '--reservation' unrecognized.` for every case — after the jobs have already been submitted and started.

## What this changes

Following the suggestion in #586 to generalize this in the style of `--extra_singularity_options`, rather than adding a single `--reservation` flag, this adds three options to `srun_fastsurfer.sh`: `--extra_slurm_options` for both jobs, plus `--extra_slurm_options_seg` and `--extra_slurm_options_surf` for one of them. Both the segmentation job and the surface job are covered, as requested in the issue. The options are inserted in front of the batch script path, because `sbatch` stops parsing options at the batch script and passes everything after it to the script instead. Option strings are split at whitespace, so several options can be given at once (documented in `--help`, which `doc/scripts/SLURM.md` renders). The cleanup and copy helper jobs are deliberately left alone, matching how `--partition` behaves today.

```console
$ ./srun_fastsurfer.sh ... --extra_slurm_options_seg "--reservation=gpu_nodes --qos=high"
sbatch --parsable --mem=7G --cpus-per-task=16 -J FastSurfer-Seg- --time=20 -o WORK/logs/seg_%A_%a.log --gpus-per-task=1 --reservation=gpu_nodes --qos=high WORK/scripts/slurm_cmd_seg.sh
sbatch --parsable --mem-per-cpu=3G --cpus-per-task=2 --ntasks=1 --nodes=1-1 --hint=nomultithread --depend=afterok:SEG_JOB_ID -J FastSurfer-Surf- -o WORK/logs/surf_%A_%a.log WORK/scripts/slurm_cmd_surf.sh
```

The general options go before the job-specific ones. So when you give an option in both, the job-specific value comes last on the `sbatch` line and takes precedence:

```console
$ ./srun_fastsurfer.sh ... --extra_slurm_options "--qos=normal" --extra_slurm_options_seg "--qos=high"
sbatch --parsable --mem=7G --cpus-per-task=16 -J FastSurfer-Seg- --time=20 -o WORK/logs/seg_%A_%a.log --gpus-per-task=1 --qos=normal --qos=high WORK/scripts/slurm_cmd_seg.sh
sbatch --parsable --mem-per-cpu=3G --cpus-per-task=2 --ntasks=1 --nodes=1-1 --hint=nomultithread --depend=afterok:SEG_JOB_ID -J FastSurfer-Surf- -o WORK/logs/surf_%A_%a.log --qos=normal WORK/scripts/slurm_cmd_surf.sh
```

@dkuegler added two related changes during review:

- `--gpus-per-task=1` now also goes before the batch script path (56e8d50). On `dev` the script appends it after `slurm_cmd_seg.sh` (see the first `sbatch` line under *What was wrong*), so `sbatch` passes it to the script as an argument and the segmentation job requests no GPU.
- `--extra_slurm_options*` and `--extra_singularity_options*` accept repeats, and the values accumulate (35ed9f1). On `dev`, a repeated `--extra_singularity_options*` keeps only its last value.

## How it was verified

`srun_fastsurfer.sh --dry` prints the `sbatch` commands and submits no jobs. We ran it in a Debian container against `dev` at `2e5604f` (its output matches the `ef190b9` lines above) and against this branch, and the `sbatch` lines in this description come from those runs. @dkuegler removed the pytest suite from the first version of this PR, because GitHub Actions cannot test SLURM interactions.


srun_fastsurfer.sh
```diff
@@ -244,6 +256,9 @@ case $key in
     ;;
   --partition_seg) partition_seg="$1" ; shift ;;
   --partition_surf) partition_surf="$1" ; shift ;;
+  --extra_slurm_options) extra_slurm_options=$1 ; shift ;;
+  --extra_slurm_options_seg) extra_slurm_options_seg=$1 ; shift ;;
+  --extra_slurm_options_surf) extra_slurm_options_surf=$1 ; shift ;;
```
dkuegler: This discards previous arguments.
```suggestion
  --extra_slurm_options) extra_slurm_options="$extra_slurm_options $1" ; shift ;;
  --extra_slurm_options_seg) extra_slurm_options_seg="$extra_slurm_options_seg $1" ; shift ;;
  --extra_slurm_options_surf) extra_slurm_options_surf="$extra_slurm_options_surf $1" ; shift ;;
```

test/scripts/test_srun_fastsurfer.py
dkuegler, line 1: We are not looking to add test scripts for main interface scripts at this point.

srun_fastsurfer.sh
```diff
@@ -154,6 +159,13 @@ SLURM-related options:
 --partition_seg <comma-separated-list-of-partitions>, and
 --partition_surf <...list>: partition(s) to schedule all or only segmentation or surface reconstruction, respectively:
   It is recommended to select nodes/partitions with GPUs for segmentation. default: slurm default partition
+--extra_slurm_options <sbatch options>,
+--extra_slurm_options_seg <sbatch options>, and
+--extra_slurm_options_surf <sbatch options>: Extra options passed to the sbatch calls of all or only the
+  segmentation or the surface reconstruction job, respectively, needs to be double quoted, e.g.
```
dkuegler: This also adds a precedence, the general options are passed before the individual options. That should probably be documented (minimally) here.

dkuegler (changes requested):
This PR looks nice. Thanks for the contribution.

I have a couple of feedback points, maybe you can rebase and address those issues?

dkuegler (approved):
Resolved comments. Probably squash-merge to drop the extra text script.

srun_fastsurfer.sh
```diff
@@ -70,6 +73,8 @@ srun_fastsurfer.sh [--data <directory to search images>]
     [--num_cases_per_task <number>] [--cpu_only] [--num_cpus_per_task <number of cpus to allocate for seg>]
     [--time_seg <timelimit>] [--time_surf <timelimit>] [--mem_seg <number (GB)>] [--mem_surf <number (GB)>]
     [--partition <slurm partition>] [--partition_seg <slurm partition>] [--partition_surf <slurm partition>]
+    [--extra_slurm_options <sbatch options>] [--extra_slurm_options_seg <sbatch options>]
+    [--extra_slurm_options_surf <sbatch options>]
```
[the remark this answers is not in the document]
dkuegler: Test suite was removed because slurm interactions cannot be tested in GitHub actions.

srun_fastsurfer.sh
```diff
@@ -577,7 +598,7 @@ then
   seg_slurm_sched=("--mem=${mem_seg}G" "--cpus-per-task=$num_cpus_per_task" -J "FastSurfer-Seg-$USER"
                    --time=$((timelimit_seg * real_num_cases_per_task + 5))
                    "${slurm_partition[@]}" "${slurm_email[@]}" "${jobarray_option[@]}"
-                   -o "$hpc_work/logs/seg_%A_%a.log" "$seg_cmd_filename")
+                   -o "$hpc_work/logs/seg_%A_%a.log" "${slurm_extra_seg[@]}" "$seg_cmd_filename")
   if [[ "$cpu_only" == "true" ]] ; then debug "Schedule SLURM job without gpu"
   else seg_slurm_sched+=(--gpus-per-task=1)
   fi
```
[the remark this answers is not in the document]
dkuegler: This is a genuine bug, interesting this never raised an error :-/

dkuegler:
> ### 🔵 Needs a closer look
> 
> The segmentation GPU request remains positioned after the batch-script path and is therefore not parsed by `sbatch`.

Failed to push commit :-/

eastagiletracker:
Thanks for the rebase and for fixing the GPU ordering along the way. We updated the description to match the final change, so squash-merge works for us whenever you're ready.

PR #854, opened 2026-08-12
Accept voxel sizes within float noise of 1mm (e.g. 1.0000001)

m-reuter:
recon-surf forwards the voxel size read from a (possibly noisy) header as an
explicit --vox_size, so a 1mm image with 1.0000001 in the header was rejected by
the <= 1.0 cap in both the CLI arg validator and conformed_vox_img_size. Round
numeric voxel sizes to vox_eps precision before the cap so 1.0000001 is treated
as 1.0, while genuinely larger sizes (e.g. 1.1) are still rejected.

Addresses comment in #850

### DIFF
--- a/FastSurferCNN/utils/arg_types.py
+++ b/FastSurferCNN/utils/arg_types.py
@@ -164,7 +164,9 @@ def float_gt_zero_and_le_one(a: str | float) -> float | None:
     """
     if a is None or isinstance(a, str) and a.lower() in ["none", "infinity"]:
         return None
-    a_float = float(a)
+    # round to 1e-4 precision (the default vox_eps used in conform) so that float noise in a voxel
+    # size read from a header, such as 1.0000001 from a noisy 1mm image, is accepted as 1.0
+    a_float = round(float(a), 4)
     if 0.0 < a_float <= 1.0:
         return a_float
     else:
--- a/FastSurferCNN/utils/arg_types.py
+++ b/FastSurferCNN/utils/arg_types.py
@@ -95,19 +95,25 @@ def vox_size(a: str | float | None) -> VoxSizeOption:
     str or float or None
         If 'auto' or 'min' is provided, it returns a string('auto' or 'min').
         If a valid voxel size (between 0 and 1) is provided, it returns a float.
-        If 'any' or 'keep', it returns None.
+        If 'any', 'keep', 'none' or 'infinity', it returns None.
 
     Raises
     ------
     ValueError
         If the argument is not "min", "auto" or convertible to a float between 0 and 1.
     """
-    if a is None or isinstance(a, str) and a.lower() in ["any", "keep"]:
+    if a is None or isinstance(a, str) and a.lower() in ["any", "keep", "none", "infinity"]:
         return None
     if isinstance(a, str) and a.lower() in ["auto", "min"]:
         return "min"
     try:
-        return float_gt_zero_and_le_one(a)
+        a_float = float(a)
+        # voxel sizes are typically read from a float32 header, so a nominal 1mm image can report
+        # 1.0000001 and would be rejected by the 1mm cap; snap noise below vox_eps (the tolerance
+        # conform compares voxel sizes with) to exactly 1.0, while 1.1 is still rejected
+        if abs(a_float - 1.0) <= 1e-4:
+            a_float = 1.0
+        return float_gt_zero_and_le_one(a_float)
     except ValueError as e:
         raise ValueError(e.args[0] + " Additionally, vox_size may be 'min'.") from None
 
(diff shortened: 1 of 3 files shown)


sRassmann:
The issue seems to persist, as the scripts still exists since an invalid argument for `--vox_size` is passed:

```
USING NEW CONFORM    # manually mounted conform.py script
[INFO: conform.py: 1729]: Image Conform Parameters:
[INFO: conform.py: 1729]: 
[INFO: conform.py: 1729]: - verbosity: True
[INFO: conform.py: 1729]: - input volume: /output/mts_MTS00071/mri/orig.mgz
[INFO: conform.py: 1729]: - check only: True
[INFO: conform.py: 1729]: - dtype: uint8
[INFO: conform.py: 1729]: - voxel size: min
[INFO: conform.py: 1729]: - round voxel size to 1mm if > threshold: None
[INFO: conform.py: 1729]: - image size: auto
[INFO: conform.py: 1729]: - affine orientation: lia
[INFO: conform.py: 1729]: - log: stdout only
[INFO: conform.py: 1934]: Reading input: /output/mts_MTS00071/mri/orig.mgz ...
[INFO: conform.py: 1453]: A conformed image must satisfy the following criteria:
[INFO: conform.py: 1457]:  - Number of Dimensions 3        : GOOD image ndim 3
[INFO: conform.py: 1457]:  - Voxel Size min=1.0x1.0x1.0    : GOOD image 1.0000001x0.99999994x1.0
[INFO: conform.py: 1457]:  - Dimensions auto=256x256x256   : GOOD image dimensions (np.int32(256), np.int32(256), np.int32(256))
[INFO: conform.py: 1457]:  - Orientation LIA               : GOOD affine=[[-1. -0. -0.] [-0. 0. 1.] [-0. -1. 0.]] => LIA
[INFO: conform.py: 1457]:  - Dtype uint8                   : GOOD dtype uint8
[INFO: conform.py: 1969]: Input /output/mts_MTS00071/mri/orig.mgz is already conformed! Exiting.

@#@FSTIME  2026:08:13:12:34:44 FastSurferCNN/data_loader/conform.py N 6 e 2.12 S 0.21 U 1.94 P 101% M 408248 F 2 R 75833 W 0 c 21 w 2423 I 11436 O 0L 9.12 6.17 5.61
@#@FSLOADPOST 2026:08:13:12:34:46 FastSurferCNN/data_loader/conform.py N 6
python3 -s /fastsurfer/FastSurferCNN/data_loader/conform.py -i /output/mts_MTS00071/mri/aparc.DKTatlas+aseg.deep.mgz --check_only --vox_size 1.0000001 --dtype any --verbose
@#@FSLOADPRE 2026:08:13:12:34:47 FastSurferCNN/data_loader/conform.py N 8 L 9.35 6.26 5.64
usage: 
Script to conform an MRI brain image to UCHAR, RAS orientation, 
and 1mm or minimal isotropic voxels

USAGE:
conform.py  -i <input> -o <output> <options>
OR
conform.py  -i <input> --check_only <options>
Dependencies:
    Python 3.10+
    Numpy
    https://www.numpy.org
    Nibabel to read and write FreeSurfer data
    https://nipy.org/nibabel/
Original Author: Martin Reuter
Modified by: David Kügler
Date: May-12-2025
conform.py: error: argument --vox_size: invalid vox_size value: '1.0000001'
```

FastSurferCNN/utils/arg_types.py
```diff
@@ -164,7 +164,9 @@ def float_gt_zero_and_le_one(a: str | float) -> float | None:
     """
     if a is None or isinstance(a, str) and a.lower() in ["none", "infinity"]:
         return None
-    a_float = float(a)
+    # round to 1e-4 precision (the default vox_eps used in conform) so that float noise in a voxel
+    # size read from a header, such as 1.0000001 from a noisy 1mm image, is accepted as 1.0
+    a_float = round(float(a), 4)
```
dkuegler: Rounding this generally here is not how this should work, the function is used in many places and fixing this precision level here may have unforeseen consequences.

As a fix, I recommend introducing an optional keyword argument `precision` with default value None, that will trigger the rounding if given. In the argparse call in conform.py, you can add then change to `partial(float_gt_zero_and_le_one, precision=1e-4)` which would trigger the rounding for this call, but I am not sure even that is really that great. If someone purposefully calls `--vox_size 0.99247` does he really expect rounding to 0.9925? This is extremely circumstantial and an edge-case, and yet relevant. I guess, we could add that as documentation for vox_size...

ClePol:
Is this a duplicate of https://github.com/Deep-MI/FastSurfer/pull/852 ?

m-reuter:
> Is this a duplicate of #852 ?

Yes, but a different solution as it also addresses cases where the code is called from within python, not only the CLI. 

Will compare the two solutions in more detail later, and also needs to address the invalid Vox size flag.

m-reuter:
@sRassmann thanks for testing. 
The fix needed a second file (`arg_types.py`) that you did not mount into your image. That is why it failed for you (only replacing `conform.py` does not fix the issue). Please test again mounting both files. I tried here with synthetic images and `--vox_size 1.0000001`. 

Two follow-ups since your report: I added regression tests (`test/image/test_vox_size_noise.py`) covering the argparse type, the parsed CLI call, and  `conformed_vox_img_size` . There was no test at this level before, which is why CI stayed green. And I adopted #852 placement, applying the tolerance inside ` vox_size()`  instead of the shared  `float_gt_zero_and_le_one()` , so ` --robust ` and  `--conform_to_1mm_threshold`  keep their strict validation.

sRassmann:
You're right, sorry for the confusion. With the fixed version mounted, the checks are passed and the images process as expected.

PR #869, opened 2026-09-04
Neurolit version bump

ClePol:
This bumps the neurolit dependency to the just-released version 0.7.0.

This version:
- has some FastSurfer compatibility fixes
- runs much faster only taking a few minutes (around 2 min for medium-sized lesions, compared to 30-60 min for the previous versions)
- has some minor bugfixes
- faster and more robust checkpoint downloading
- is tested with the validation pipeline (in combination with FastSurfer)

The only compatibility issue that I could see arise is that this depends on a fairly new Monai version - testing didn't show any issues though.

Since this is now tested thoroughly, I also removed the experimental wording from the documentation.


ClePol:
Converted to draft because the new Monai version could indeed break things because it requires pytorch<=2.8 and that may not be compatible with the docker images supporting old CUDA drivers. We should discuss this offline

Issue #857, opened 2026-08-24
Unable to Install FastSurfer on MacBook M2


so-hard-to-name:
Bug Report
==========

Description
-----------
Install FastSurfer on MacBook Air M2 failed. 
1. Installer stuck in "Running package scripts" for long time
2. Finally, got text:
"The installation failed. The Installer encountered an error that caused the installation to fail. Contact the software manufacturer for assistance."

Also I tried removing FreeSurfer first, this happened as well.

Expected Behavior
-----------------
Expect a successful installation

Steps to Reproduce
------------------
1. Install Python 3.10+ and Bash
2. Download installation package, FastSurfer254-macos-darwin_arm64.pkg file.
3. Double click the .pkg file to start the installation.
4. Agree the license agreement and go to the Installation stage
5. -> writing files and other steps went well, but in "running package scripts", it stucked for long time

Log Files / Screenshots
-----------------------

<img width="621" height="445" alt="Image" src="https://github.com/user-attachments/assets/2cb177ac-5fa4-4596-a5cc-c8547e39e874" />
<img width="620" height="451" alt="Image" src="https://github.com/user-attachments/assets/1699deb9-ee23-44e3-bb27-9f4dec5e51fb" />

Environment
-----------

- **Installation type**: package
- FastSurfer Version: 2.5.4
- FreeSurfer Version: this situation happened both when FreeSurfer is installed and not installed 
- OS: Mac
- ARM64

Additional Context
------------------
Add any other context and comments about the problem here.

m-reuter:
@so-hard-to-name thanks for reporting, I just replicated this on an M1 Mac. Actually the error message is wrong as it most likely went all the way through. It can take a long time as it downloads the current checkpoints. At the end it catches a harmless exit 1 and then reports failure even though everything is installed properly. If you start FastSurfer2.5.4 you should get the console and be able to run things. This PR #859 is designed to fix this in future releases (but still needs testing).

PR #877, opened 2026-09-10
Fix macOS package download link

idossha:
The macOS package link returns 404 because release assets include the version in their filename (currently `FastSurfer254-macos-darwin_arm64.pkg`). Link to the latest release's Assets section instead, currently [v2.5.4](https://github.com/Deep-MI/FastSurfer/releases/tag/v2.5.4), so future package names do not break the instructions.

Verified: old link returns 404; replacement resolves successfully; Ruff, codespell and all 77 contributor shell tests pass.


m-reuter:
The documentation is in line with dev. In dev we changed the name of the assets to drop the version in the file name, so that a direct click is possible. This will only show at the next release, so currently the link is broken in the dev Documentation. I considered re-uploading a second Mac package with the new name to the latest release, but since the package installer is broken in that release anyway (see #859 and #857) I decided against it. 

The stable Documentation points to the correct location:
https://deep-mi.org/FastSurfer/stable/overview/INSTALL.html

Anyway thanks for your contribution, but will not merge. I am however considering to have users land not at the dev documentation by default but at the stable one, tied to the release to avoid such confusion in the future.

idossha:
No worries. Thanks for the input and explanation. Did not even realized I was on dev.. still new to the repo. cool stuff

Issue #882, opened 2026-09-14
UPX-packed FreeSurfer binaries in the cuda-v2.5.x images are SIGKILLed at launch on hosts running CrowdStrike Falcon


Arshya-Guru:
Bug Report
==========

Description
-----------
The official `deepmi/fastsurfer:cuda-v2.5.4` image (and 2.5.1–2.5.3, which are packed identically) ships every FreeSurfer binary in `/opt/freesurfer/bin` compressed with **UPX 4.22** (`tools/build/install_fs_pruned.sh --upx`, hard-coded in `tools/Docker/Dockerfile` line 226, ubuntu:24.04 base). On our HPC cluster, whose nodes run the **CrowdStrike Falcon** EDR sensor, the UPX self-extraction stub is SIGKILLed by the sensor at launch for some of these binaries — `mri_add_xform_to_header` essentially every time, `mris_fix_topology` most of the time. As a result every `recon-surf.sh` run (cross-sectional and longitudinal) dies at the Talairach step or at topology fixing.

Three things make this hard to diagnose for users:

1. The only message is `Command terminated by signal 9` / `Killed`, which every existing issue on the tracker (#615, #503, #397, #383 …) attributes to out-of-memory. Adding RAM does nothing.
2. The next command then fails with `mghRead(.../nu.mgz): could not open file`, so users report a "cannot open file" error as a second, unrelated bug.
3. `long_fastsurfer.sh` (and `run_fastsurfer.sh`) do not propagate the failure: after all three recon-surf stages die, the log still ends with `Full longitudinal processing for sub-092 finished!` and the process exits 0.

The 2.4.2 image (ubuntu:22.04, UPX 3.96) is not affected on the same hosts; `mri_add_xform_to_header` is not even packed there (UPX 3.96 apparently skipped it). Unpacking all 67 binaries in place with `upx -d` fully resolves the problem; the unpacked payload of the 2.5.4 binary is **byte-identical** to the 2.4.2 copy (md5 `75edc6f92f224262a28aa185a4373bb7`, FreeSurfer build `freesurfer-linux-ubuntu22_x86_64-7.4.1-20230614-7eb8460`), so the packing is the only difference.

Expected Behavior
-----------------
[...]

PR #883, opened 2026-09-14
Do not UPX-pack FreeSurfer binaries for Docker

Arshya-Guru:
Since 2.5.4 the image is built on ubuntu:24.04 (UPX 4.22). On hosts running EDR software such as CrowdStrike Falcon, the UPX self-extraction stub is SIGKILLed at launch for several FreeSurfer binaries (mri_add_xform_to_header,
mris_fix_topology), so every recon-surf run fails with what looks like an OOM.
The size saving from packing is ~7 MB on a ~5.5 GB image.


m-reuter:
Thanks for the PR and issue. We should drop upx then, but as copilot says, this needs a little more work (on our side) to update caches etc. Also this does not fix #882 completely as it also found missing error propagation in some scripts that need to be checked separately.

m-reuter:
Merging. Note this only affects builds that compile FreeSurfer from source. Official images copy it from the prebuilt  deepmi/fastsurfer-build:freesurfer741 , which is still packed, so #882 stays open until that is rebuilt.

